PR #51, opened 2015-03-31
add --dry-run arg to rosdistro_reformat

mintar:


scripts/rosdistro_reformat
```diff
@@ -37,7 +37,7 @@ import argparse
 import os
 import sys
 
-from rosdistro.verify import reformat_files
+from rosdistro.verify import reformat_files, verify_files_identical
```
dirk-thomas: Please use separate imports following PEP8.

mintar: Well, PEP8 says it's okay:

> Imports should usually be on separate lines, e.g.:
> 
> Yes:
>    import os
>    import sys
> 
> No:  import sys, os
> 
> It's okay to say this though:
> 
> from subprocess import Popen, PIPE

But I can of course change it, whatever you prefer!

dirk-thomas: Thanks for clarifying. Can you please still split the lines :wink: Thanks.

mintar: Done!

dirk-thomas:
Thank you!

PR #52, opened 2015-03-31
fix the return code of rosdistro_reformat

mintar:
rosdistro_reformat now returns 0 if no files have been changed and 1 otherwise


dirk-thomas:
Thanks for the patch.

PR #53, opened 2015-07-14
Proper check of debian version according to Debian Versioning Policy

v-lopez:
According to https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version : 
> debian_revision
>    This part of the version number specifies the version of the Debian package based on the upstream version. It may contain only alphanumerics and the characters + . ~ (plus, full stop, tilde) and is compared in the same way as the upstream_version is. 

This fix ensures that versions that are compliant with Debian standard are accepted by the cache generators.

### DIFF
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -95,7 +95,7 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
             errors.append('%s: invalid package.xml file for package "%s": %s' % (dist_name, pkg_name, e))
             continue
         # check that version numbers match (at least without deb inc)
-        if not re.match('^%s-\d+$' % re.escape(pkg.version), repo.version):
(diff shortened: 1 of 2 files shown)


wjwwood:
Can you give us a little context as to why you think this change needs to be made?

I'd point out that our version scheme is not the same as Debian's: http://www.ros.org/reps/rep-0140.html#version

v-lopez:
In our use case, we append some extra information to distinguish between two debians built from the same version and same build number but built against different repositories.(one built against the last version available and the other against the most stable version of the software).

wjwwood:
Hmm, it's not clear to me if this is ok to allow or not (it might affect more of our tools). I'd say I'm +0 to merge this. I'll let @tfoote or @dirk-thomas weigh in.

dirk-thomas:
I don't see a problem with relaxing this check. I will let @tfoote merge once he has reviewed it.

Afterwards you can go ahead and try releasing a package with that suffix and we will find out if there are any problems down the road which need to be addressed too :wink:

tfoote:
I don't see a reason why this would break. Please do test this to verify it works.

Issue #48, opened 2015-02-19
invalid urls to .inv files


dirk-thomas:
https://github.com/ros-infrastructure/rosdistro/blob/master/doc/conf.py#L219-L225

PR #54, opened 2015-12-09
fix inv urls

dirk-thomas:
Addresses #48. Should fix the 404s in http://jenkins.ros.org/job/_doc_generation/


wjwwood:
+1

dirk-thomas:
Merging anyway since the failing test with Python 3.2 and coverage is a known problem

PR #57, opened 2015-12-20
Support url query parameters

otamachan:
Hi,

I'm using Atlassian Stash to manage my own ``rosdsitro``.
Unfortunately, in order to get a raw content of a file from Stash, I need to add query parameters like ``raw`` and ``at=``.
(http://stackoverflow.com/questions/14063008/get-raw-file-content-using-stash-rest-api)
Although there is already an issue, nothing has seemed to progress so far(https://jira.atlassian.com/browse/BSERV-4036) .

I would appreciate if url query parameters are supported, that is if ``index_url`` has query parameters then all the relative urls have the same queries.

Thanks.

### DIFF
--- a/src/rosdistro/__init__.py
+++ b/src/rosdistro/__init__.py
@@ -48,6 +48,10 @@ try:
 except ImportError:
     from io import BytesIO as StringIO
 import sys
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
 import yaml
 
 logger = logging.getLogger('rosdistro')
@@ -113,7 +117,8 @@ def get_index(url):
     yaml_str = load_url(url)
     data = yaml.load(yaml_str)
     base_url = os.path.dirname(url)
-    return Index(data, base_url)
+    url_parts = urlparse(url)
+    return Index(data, base_url, url_query=url_parts.query)
 
 
 ### distribution information
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -41,3 +41,34 @@ def test_get_index_v3_invalid():
         assert False
     except AssertionError:
         pass
+
+
+def test_get_index_from_http_with_query_parameters():
+    import subprocess
+    import sys
+    import time
+    url = 'http://localhost:9876/index_v3.yaml?raw&at=master'
+    # start a http server and wait
+    if sys.version_info < (3, 0, 0):
+        proc = subprocess.Popen([sys.executable, '-m', 'SimpleHTTPServer', '9876'],
+                                cwd=FILES_DIR)
+    else:
+        proc = subprocess.Popen([sys.executable, '-m', 'http.server', '9876'],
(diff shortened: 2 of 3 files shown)


wjwwood:
Code review looks ok to me, and by the way, thanks for including a test.

I'd need to see this tested with all of our tools to make sure nothing else is negatively affected, before I'd give it a +1. Not sure when we'll get around to that, as several people will be on vacation for the next few weeks.

dirk-thomas:
This changes the semantic of the public API. How should we move forward on this?

otamachan:
Thank you for your feedback.
Keeping the ``base_url`` parameter and adding an extra keyword parameter like ``query`` is acceptable?

dirk-thomas:
@otamachan Can you please give an estimate when you will update the PR? If it will be pretty soon (today or early tomorrow) we can release it with the next patch version. Otherwise it would release a patch version without it and it will go into the next one (which is uncertain when that will be).

otamachan:
＠dirk-thomas I'll update the PR from now. It will ends within an hour.

dirk-thomas:
Great. Thanks.

otamachan:
I updated the PR. Please review it.

src/rosdistro/__init__.py
```diff
@@ -49,6 +49,10 @@
     from io import BytesIO as StringIO
 import sys
 import yaml
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
```
dirk-thomas: Please move this block on line higher (above `yaml`) to keep the import order alphabetically.

test/test_index.py
```diff
@@ -41,3 +41,34 @@ def test_get_index_v3_invalid():
         assert False
     except AssertionError:
         pass
+
+
+def test_get_index_from_http_with_query_parameters():
+    import time
```
dirk-thomas: Please move this line two lines lower (below `sys`) to keep the import order alphabetically.

test/test_index.py
```diff
@@ -41,3 +41,34 @@ def test_get_index_v3_invalid():
...
+    import subprocess
+    import sys
+    url = 'http://localhost:9876/index_v3.yaml?raw&at=master'
+    # start a http server and wait
+    if sys.version_info < (3, 0, 0):
+        proc = subprocess.Popen(['python', '-m', 'SimpleHTTPServer', '9876'],
```
dirk-thomas: Instead of using `python` please use `sys.executable`.

Same below.

dirk-thomas:
Beside the small comments the patch looks good to me.

otamachan:
I'll fix soon.

otamachan:
Done. Please review it.

dirk-thomas:
Thanks for  iterating on this. +1 @wjwwood ?

wjwwood:
lgtm

PR #58, opened 2016-02-02
fix retry on socket timeout with Python 3

dirk-thomas:
Same patch as in ros-infrastructure/ros_buildfarm#168

This fix should be released asap since jobs are failing due to this: e.g. http://build.ros.org/job/Jbin_arm_uThf__rviz_python_tutorial__ubuntu_trusty_armhf__binary/3/console


tfoote:
+1

PR #59, opened 2016-02-02
retry on different http error code

dirk-thomas:
Fails some builds, e.g. http://build.ros.org/job/Irel_arm_release-status-page/2032/console


tfoote:
+1

Issue #63, opened 2016-02-20
Make manifest_provider.git more efficient


mikepurvis:
Right now for non-github repos, it has to clone the entire thing just to grab the package.xml. Per the [comment here](https://github.com/ros-infrastructure/rosdistro/blob/39455f500c8f771939e0286672547a06e4d9c233/src/rosdistro/manifest_provider/git.py#L59), this is due to a limitation with Git 1.7.9, but the oldest supported platform now is Ubuntu Trusty, which has Git 1.9.1.

So it should be possible now to change this so that it does a shallow clone of only the commit that is the tag. It's a significant savings, particularly on long-running repos which have had a lot of churn:

```
$ git clone https://github.com/ros-gbp/navigation-release.git
Cloning into 'navigation-release'...
remote: Counting objects: 67383, done.
remote: Total 67383 (delta 0), reused 0 (delta 0), pack-reused 67383
Receiving objects: 100% (67383/67383), 8.02 MiB | 1.14 MiB/s, done.
Resolving deltas: 100% (22599/22599), done.
Checking connectivity... done.
```

```
$ git clone https://github.com/ros-gbp/navigation-release.git --depth 1 --branch upstream/1.13.1
Cloning into 'navigation-release'...
remote: Counting objects: 411, done.
remote: Compressing objects: 100% (307/307), done.
remote: Total 411 (delta 105), reused 258 (delta 66), pack-reused 0
Receiving objects: 100% (411/411), 469.52 KiB | 0 bytes/s, done.
Resolving deltas: 100% (105/105), done.
Checking connectivity... done.
```

Cloning from Github for example's sake, but obviously this issue affects primarily private repos on Bitbucket, internal Gitlab instances, etc. And I may yet submit a manifest provider for Gitlab, but just this would be a simple change that could have a pretty big positive impact.

dirk-thomas:
Currently the Python tools like `python-rosdistro` are distribution agnostic so the same version gets released into all Ubuntu distros / Debian distros (listed in https://github.com/ros-infrastructure/rosdistro/blob/39455f500c8f771939e0286672547a06e4d9c233/stdeb.cfg#L6). Based on that such a change wouldn't be possible since it would break installations of ROS on older distributions which only have older versions of git.

That being said it might be reasonable to consider changing our policy and removing older distros form this set of suites so that future releases are only made into "current" Ubuntu / Debian distributions. @tfoote @wjwwood What do you think about this?

[...]

PR #64, opened 2016-02-20
Clone the exact tag when fetching package.xml.

mikepurvis:
This PR addresses #63.

Tested with Trusty's Git 1.9.1.

### DIFF
--- a/src/rosdistro/manifest_provider/git.py
+++ b/src/rosdistro/manifest_provider/git.py
@@ -31,6 +31,7 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import distutils.version
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


dirk-thomas:
See comments on #63. As an alternative the code could query the git version on keep the existing code for backward compatibility and use the faster new codepath if the git version is high enough.

mikepurvis:
I'm happy for it to go either way. I'm doubtful that anyone out there is building rosdistro caches on Ubuntu Precise, but I understand the principle. I'm happy to paste the old logic back in, I guess it's just balancing complexity vs. support. Another way would be to add another manifest provider for "git-1.7", and just allow the fallback to happen through that mechanism.

dirk-thomas:
Based on the results of the discussion summarized in https://github.com/ros-infrastructure/rosdistro/issues/63#issuecomment-187426738 rosdistro needs to maintain compatible with Ubuntu Precise. Therefore please update the PR to check for git version and only use the optimized code path if the git version is new enough. Otherwise fall back to the previous more inefficient workflow.

mikepurvis:
Aight, here we are. I verified this by building a cache containing bitbucket SSH URLs and using `ps` to observe which processes were invoked during the build. I changed the version number check to `1.9.2` instead, and then it used the legacy logic, as expected.

src/rosdistro/manifest_provider/git.py
```diff
@@ -53,35 +53,50 @@ def git_manifest_provider(_dist_name, repo, pkg_name):
...
 def _get_package_xml(url, tag):
     base = tempfile.mkdtemp('rosdistro')
     try:
-        # git 1.7.9 does not support cloning a tag directly, so doing it in two steps
         assert _git_client_executable is not None, "'git' not found"
-        cmd = [_git_client_executable, 'clone', url, base]
-        result = _run_command(cmd, base)
-        if result['returncode'] != 0:
-            raise RuntimeError('Could not clone repository "%s"' % url)
+        if _git_version_gte('1.9.1'):
```
mikepurvis: I briefly grepped through the Git changelogs to try to figure out when shallow-cloning tags became possible. It wasn't immediately obvious, so I've just made Trusty's version the changeover point.

src/rosdistro/manifest_provider/git.py
```diff
@@ -104,7 +119,7 @@ def check_remote_tag_exists(url, tag):
     return False
 
 
-def _run_command(cmd, cwd, env=None):
+def _run_command(cmd, cwd='/', env=None):
```
dirk-thomas: I don't think that a default working directory of '/'  is a good idea. Instead please make the default `None` which implies the current working directory.

src/rosdistro/manifest_provider/git.py
```diff
@@ -53,35 +53,50 @@ def git_manifest_provider(_dist_name, repo, pkg_name):
         raise RuntimeError('Unable to fetch package.xml: %s' % e)
 
 
+def _git_version_gte(version):
+    global _git_client_version
+    if not _git_client_version:
+        cmd = [_git_client_executable, '--version']
+        result = _run_command(cmd)
+        _git_client_version = result['output'].split()[-1]
+    return _git_client_version.split('.') >= version.split('.')
```
dirk-thomas: String comparison is not adequate here. It needs to compare the integer values. Otherwise "5" > "10".

Also the Python module `distutils.version` contains classes and functions to parse and compare version numbers.

mikepurvis: I had an earlier version which used `distutils.version.StrictVersion`; I'll switch back to that.

src/rosdistro/manifest_provider/git.py
mikepurvis: Sounds good, will change.

mikepurvis:
Please check now.

src/rosdistro/manifest_provider/git.py
dirk-thomas: I ran `man git clone` on other machine and in the paragraph of `--branch` it mentions something like this:

    --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

The earliest minor version where this was available is 1.8 (see https://git.kernel.org/cgit/git/git.git/tree/Documentation/git-clone.txt?h=v1.8.0#n158).

mikepurvis: Updated the version and comment accordingly.

dirk-thomas:
Looks good to me. Can you please squash the commits.

mikepurvis:
Squashed.

dirk-thomas:
Thank you!

dirk-thomas:
The new check (`StrictVersion`) breaks the `git` manifest provider (https://github.com/ros/rosdistro/pull/10930) since it might report a version number with four parts (e.g. 1.8.5.6) which does not comply with the strict version number (see https://www.python.org/dev/peps/pep-0386/#distutils).

I created #74 to address the problem.

PR #65, opened 2016-02-23
Make debian increment optional when comparing versions.

mikepurvis:
This is for cases when you want to include packages in a cache (and thus available to rosinstall_generator) without making a bloomed release of them, for example by referencing the upstream directly in the `release` stanza, eg:

```
  libg2o:
    release:
      tags:
        release: "{version}"
      url: https://github.com/mikepurvis/libg2o.git
      version: 2016.2.23
    source:
      type: git
      url: https://github.com/mikepurvis/libg2o.git
      version: indigo-devel
```

Without the proposed change, the above results in a ridiculous error message like this:

```
$ rosdistro_build_cache index.yaml                                                                                                         
Build cache for "clearpath"
- trying to use local cache
- use local file "clearpath-cache.yaml.gz"
- update cache
- fetch missing manifests
.........................................................................................................................................................................................................
.........................................................................................................................................................................................................
.........................................................................
clearpath: different version in package.xml (2016.2.23) for package "libg2o" than for the repository (2016.2.23) (after removing the debian increment)
```

### DIFF
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -95,7 +95,7 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
             errors.append('%s: invalid package.xml file for package "%s": %s' % (dist_name, pkg_name, e))
             continue
         # check that version numbers match (at least without deb inc)
-        if not re.match('^%s-[\dA-z~\+\.]+$' % re.escape(pkg.version), repo.version):
+        if not re.match('^%s(-[\dA-z~\+\.]+)?$' % re.escape(pkg.version), repo.version):
             errors.append('%s: different version in package.xml (%s) for package "%s" than for the repository (%s) (after removing the debian increment)' % (dist_name, pkg.version, pkg_name, repo.version))
 
     if not debug:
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I might be missing something but how does the release entry work if it is not a gbp repository which contains the Debian changelog / control / rules files? It seems to only contain the snapshot of the upstream code.

mikepurvis:
It can still be used to populate source workspaces from tagged source repositories.

dirk-thomas:
Isn't that what the `source` entry is good for? The build farm would very likely fail for that kind of release entry.

mikepurvis:
This is for rosdistro repos used in contexts other than **ros_buildfarm**.

The issue is that `rosdistro_build_cache` works exclusively from release repos, so the `source` entry alone is not adequate to populate the cache.

tfoote:
The urls you're putting into this file are not valid for the definition of the repo: http://www.ros.org/reps/rep-0141.html This is fixing the first error, but I expect there will be more errors downstream in different workflows if not in the rosinstall_generator workflow. This example breaks if you point to a repo with multiple packages. 

It would be better to optionally extend the cache building to explicitly know how to walk source repos. And then with the extended cache the rosinstall_generator could take advantage of the extended cache and fallback to upstream sources.

mikepurvis:
Mmkay, I'll reconsider my choices.

PR #69, opened 2016-03-08
New script: rosdistro_freeze_source

mikepurvis:
This script offers new functionality for users looking to create reproducible snapshots of _upstream source_, across an entire distro. This tool is a building block in a larger intended workflow, which is:

- freeze the distro.
- generate a cache specific to the frozen state, update index.yaml to point to it.
- tag the distro.
- use rosinstall_generator, ros_buildfarm, etc against the tagged index.yaml
- the frozen state may be "awakened" by committing the tag as a branch, and transforming select source versions _back_ to development branches, allowing development and source-based testing to resume from the point in the past.

Example uses:

```
$ git clone https://github.com/clearpathrobotics/public-rosdistro.git
$ cd public-rosdistro

# Change version pointers to hash from devel branches
$ rosdistro_freeze_source index.yaml indigo  # Implies --source-version
$ git diff
$ git checkout .

# Change version pointers to hash from latest release
$ rosdistro_freeze_source index.yaml indigo --release-version
$ git diff
$ git checkout .

# Change version pointers to tag name from latest release
rosdistro_freeze_source index.yaml indigo --release-tag
git diff

# Options
$ rosdistro_freeze_source -h
usage: rosdistro_freeze_source [-h]
                               [--source-version | --release-version | --release-tag]
                               index [dist_names [dist_names ...]]

Freeze a rosdistro's source branch versions to hashes or tags.

positional arguments:
  index              Path to a local index.yaml file.
  dist_names         The names of the distributions (default: all)

optional arguments:
  -h, --help         show this help message and exit
  --source-version   Freeze to the hash of current source branch (default).
  --release-version  Freeze to the hash of current release tag.
  --release-tag      Freeze to name of current release tag.
```

@dirk-thomas

### DIFF
--- a/scripts/rosdistro_freeze_source
+++ b/scripts/rosdistro_freeze_source
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2016, Clearpath Robotics
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of Open Source Robotics Foundation, Inc. nor
+#    the names of its contributors may be used to endorse or promote
+#    products derived from this software without specific prior
+#    written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import argparse
+import os.path
+import sys
+import yaml
+
+from rosdistro import get_distribution
+from rosdistro.index import Index
+from rosdistro.freeze_source import freeze_distribution_sources, CONCURRENT_DEFAULT
+from rosdistro.writer import yaml_from_distribution_file
+
+
+def parse_args(args=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='''
+        Freeze a rosdistro\'s source branch versions to hashes or tags. If neither --release-version
+        nor --release-tag are specified, the hashes of the current devel branches are used.
+        ''')
+    parser.add_argument('index', type=argparse.FileType(),
+                        help='Path to a local index.yaml file.')
+    parser.add_argument('dist_names', nargs='*', help='The names of the distributions (default: all)')
+    parser.add_argument('-j', '--jobs', type=int, default=CONCURRENT_DEFAULT,
+                        help='How many worker threads to use.')
+    parser.add_argument('-q', '--quiet', action="store_true",
+                        help='Suppress updating status bar (for script/CI usage).')
+
+    mode = parser.add_mutually_exclusive_group()
+    mode.add_argument('--release-version', action="store_true",
+                      help='Freeze to the hash of current release tag.')
+    mode.add_argument('--release-tag', action="store_true",
+                      help='Freeze to name of current release tag.')
+    return parser.parse_args(args)
+
+
+def main():
+    args = parse_args()
+    index = Index(yaml.safe_load(args.index),
+            'file://%s' % os.path.dirname(os.path.abspath(args.index.name)))
+
+    if not args.dist_names:
+        args.dist_names = sorted(index.distributions.keys())
+
+    for dist_name in args.dist_names:
+        dist = get_distribution(index, dist_name)
+        freeze_distribution_sources(dist, release_version=args.release_version, release_tag=args.release_tag,
+                                    concurrent_ops=args.jobs, quiet=args.quiet)
+        dist_file_local = index.distributions[dist_name]['distribution'].split('://')[1]
+        with open(dist_file_local, 'w') as f:
+            f.write(yaml_from_distribution_file(dist))
+
+
+if __name__ == '__main__':
+    main()
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ setup(
     scripts=[
 #        'scripts/rosdistro',
         'scripts/rosdistro_build_cache',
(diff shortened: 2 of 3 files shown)


scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,107 @@
...
+
+try:
+    import queue
+except ImportError:
+    import Queue as queue
+
+from rosdistro import get_distribution
+from rosdistro.writer import yaml_from_distribution_file
+
+CONCURRENT_OPERATIONS=16
```
mikepurvis: Unsure if it's worth parameterizing this. 16 seems like a reasonable balance between parallelism and not hammering your git server (or Github).

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,107 @@
...
+    parser.add_argument('dist_names', nargs='*',
+        help='The names of the distributions (default: all)'
+    )
+    parser.add_argument('--release-version', action="store_true", default=False,
+        help='Freeze to the hash of current release tag (default: the source branch hash)'
+    )
+    return parser.parse_args(args)
+
+
+def freeze_distribution_sources(dist, release_version=False):
```
mikepurvis: Happy to move this to its own module if that would be preferred.

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,107 @@
...
+            work_queue.task_done()
+
+        except queue.Empty:
+            break
+
+
+def main():
+    args = parse_args()
+    with open(args.index) as f:
+       index = Index(yaml.safe_load(f), 'file://' + os.path.dirname(os.path.abspath(args.index)))
```
mikepurvis: Pasting on the `file://` feels hackish, but I couldn't find a better way to handle it.

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
```
mikepurvis: Is there a preferred copyright block for third-party contributions to OSRF packages?

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,107 @@
...
+                hash, ref = line.split('\t', 1)
+                if ref.endswith(freeze_version):
+                    source_repo.version = hash
+                    break
+
+            work_queue.task_done()
+
+        except subprocess.CalledProcessError:
+            print("Non-zero return code for: %s" % ' '.join(cmd))
+            work_queue.task_done()
```
mikepurvis: Open to suggestions on a more graceful means of handling errors.

scripts/rosdistro_freeze_source
tfoote: For legacy packages matching BSD headers ala https://github.com/ros-infrastructure/rosdistro/blob/master/src/rosdistro/distribution.py#L1-L32, for new work Apache 2.0 https://www.apache.org/licenses/LICENSE-2.0

You can just put your own name(company name if it's on company time) and the year.

scripts/rosdistro_freeze_source
tfoote: There are more robust ways to do it with urlparse etc but it's effectively the same. Of course if you want to do it python2/3 it's going to need to be conditional ;-) https://stackoverflow.com/questions/11687478/convert-a-filename-to-a-file-url

mikepurvis:
I don't believe the tests are failing due to anything actually in this change— seems like it's some kind of issue with a flaky test dependency.

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument('index', help='Path to a local index.yaml file.')
+    parser.add_argument('dist_names', nargs='*', help='The names of the distributions (default: all)')
+
+    mode = parser.add_mutually_exclusive_group()
+    mode.add_argument('--source-version', action="store_true",
+                      help='Freeze to the hash of current source branch (default).')
+    mode.add_argument('--release-version', action="store_true",
+                      help='Freeze to the hash of current release tag.')
+    mode.add_argument('--release-tag', action="store_true",
+                      help='Freeze to name of current release tag.')
```
mikepurvis: Another way to do this is:

```
mode = parser.add_mutually_exclusive_group(required=True)
```

And then one of the options is always explicitly required.

scripts/rosdistro_freeze_source
dirk-thomas: I think it would be good to make it a command line argument. Even if nobody will ever change it it nicely makes that detail visible to the user. But you call - since I might not use the script anyway :wink:

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,81 @@
...
+from rosdistro import get_distribution
+from rosdistro.index import Index
+from rosdistro.freeze_source import freeze_distribution_sources
+from rosdistro.writer import yaml_from_distribution_file
+
+
+def parse_args(args=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Freeze a rosdistro\'s source branch versions to hashes or tags.')
+    parser.add_argument('index', help='Path to a local index.yaml file.')
```
dirk-thomas: Maybe add a check that the passed argument is actually an existing file. `argparse` can do that for you.

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument('dist_names', nargs='*', help='The names of the distributions (default: all)')
+
+    mode = parser.add_mutually_exclusive_group()
+    mode.add_argument('--source-version', action="store_true",
+                      help='Freeze to the hash of current source branch (default).')
```
dirk-thomas: This argument doesn't seem to be used right now - I guess it is only there for the default? Should the default be described in the group description instead?

scripts/rosdistro_freeze_source
```diff
@@ -0,0 +1,81 @@
...
+    with open(args.index) as f:
+        index = Index(yaml.safe_load(f), 'file://' +
+                      os.path.dirname(os.path.abspath(args.index)))
+
+    if not args.dist_names:
+        args.dist_names = sorted(index.distributions.keys())
+
+    for dist_name in args.dist_names:
+        dist = get_distribution(index, dist_name)
+        freeze_distribution_sources(dist, args.release_version, args.release_tag)
```
dirk-thomas: Please pass keyword arguments explicitly as keyword arguments.

setup.py
```diff
@@ -19,6 +19,7 @@
     scripts=[
 #        'scripts/rosdistro',
         'scripts/rosdistro_build_cache',
+        'scripts/rosdistro_freeze_source',
```
dirk-thomas: This file should have the executable flag set.

src/rosdistro/freeze_source.py
```diff
@@ -0,0 +1,103 @@
...
+                    if freeze_to_tag and ref.startswith('refs/tags/'):
+                        source_repo.version = ref.split('refs/tags/')[1]
+                    else:
+                        source_repo.version = hash
+                    break
+
+            work_queue.task_done()
+
+        except subprocess.CalledProcessError:
+            print("Non-zero return code for: %s" % ' '.join(cmd))
```
dirk-thomas: This message should go to `stderr`.

scripts/rosdistro_freeze_source
mikepurvis: No group description supported: https://docs.python.org/2/library/argparse.html#mutual-exclusion

I've dropped the arg and augmented the ArgumentParser's description.

scripts/rosdistro_freeze_source
mikepurvis: Done.

setup.py
mikepurvis: Done

src/rosdistro/freeze_source.py
```diff
@@ -0,0 +1,106 @@
...
+                    if freeze_to_tag and ref.startswith('refs/tags/'):
+                        source_repo.version = ref.split('refs/tags/')[1]
+                    else:
+                        source_repo.version = hash
+                    break
+
+            work_queue.task_done()
+
+        except subprocess.CalledProcessError:
+            sys.stderr.write("Non-zero return code for: %s" % ' '.join(cmd))
```
dirk-thomas: This line lacks a trailing newline. Since it is very common to miss when using `stderr` directly I recommend using:

    from __future__ import print_function

    print('Non-zero return code for: %s' % ' '.join(cmd), file=sys.stderr)

mikepurvis: Done.

dirk-thomas:
LGTM. Can you please squash the commits before the merge.

Do you think this is ready to also be released or do you imagine more features / changes soon?

mikepurvis:
Squashed. This is good to release.

dirk-thomas:
Thank you!

PR #70, opened 2016-03-09
print package name if manifest was updated

dirk-thomas:
This is necessary for ros-infrastructure/ros_buildfarm#235.

Once this is merged I will do a new patch release.


wjwwood:
I can't speak to how this works with a solution to https://github.com/ros-infrastructure/ros_buildfarm/issues/235, but the code changes in this PR lgtm.

dirk-thomas:
I have added the package version to the output to provide better output for the user in case of two releases close after each other.

wjwwood:
Looks fine (I restarted CI which had a connection failed error).

PR #71, opened 2016-03-10
print repo name and section of doc or source entry changes

dirk-thomas:
Similar to #70 but for the doc and source entries. Currently the build farm doesn't support reconfiguring a subset of devel and doc jobs but with this information the same is possible as for the release reconfiguration.

Once this is merged I will do a new patch release.


wjwwood:
lgtm

PR #72, opened 2016-03-23
Fix for cache updating.

mikepurvis:
This fixes a nasty little bug I ran into where my cache got corrupted and then wasn't getting the distribution portion updated any longer:

```
$ rosdistro_build_cache index.yaml indigo
Build cache for "indigo"
- trying to use local cache
- use local file "indigo-cache.yaml.gz"
- failed to fetch old cache: string indices must be integers, not str
- update cache
- fetch missing manifests
............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
- write cache file "indigo-cache.yaml"
```

### DIFF
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -130,14 +130,15 @@ def _get_cached_distribution(index, dist_name, preclean=False, ignore_local=Fals
                 print('- trying to fetch cache')
                 # get distribution cache
                 cache = get_distribution_cache(index, dist_name)
-            # get current distribution file
-            rel_file_data = _get_dist_file_data(index, dist_name, 'distribution')
-            # update cache with current distribution file
-            cache.update_distribution(rel_file_data)
     except Exception as e:
         print('- failed to fetch old cache: %s' % e)
+
     if cache:
         print('- update cache')
+        # get current distribution file
+        rel_file_data = _get_dist_file_data(index, dist_name, 'distribution')
+        # update cache with current distribution file
+        cache.update_distribution(rel_file_data)
     else:
         print('- build cache from scratch')
         # get empty cache with distribution file
--- a/src/rosdistro/release_cache_generator.py
+++ b/src/rosdistro/release_cache_generator.py
@@ -129,14 +129,15 @@ def _get_cached_release(index, dist_name, preclean=False):
                 print('- trying to fetch cache')
                 # get release cache
                 cache = get_release_cache(index, dist_name)
-            # get current release file
-            rel_file_data = _get_dist_file_data(index, dist_name, 'release')
-            # update cache with current release file
-            cache.update_distribution(rel_file_data)
     except:
         print('- failed to fetch old cache')
+
     if cache:
         print('- update cache')
+        # get current release file
+        rel_file_data = _get_dist_file_data(index, dist_name, 'release')
+        # update cache with current release file
+        cache.update_distribution(rel_file_data)
     else:
         print('- build cache from scratch')
         # get empty cache with distribution file


dirk-thomas:
If I read the core correctly the exception which was catched (`string indices must be integers, not str`) must have happened in the two lines you moved in this patch. Otherwise I don't see how `cache` would be set in case of the exception block being triggered. Doesn't that imply that moving the code as in this patch would still raise the same exception - but then it is not being handled anymore?

Can you describe what exactly led to this problem in the first place?

mikepurvis:
I'll observe for when it happens again and report back with more specifics. Another potential fix would be to simply put `cache = None` in the except block, so that it builds from scratch in the case where either loading or updating the previous cache fails.

dirk-thomas:
What about the following:

* the fetching of the cache was successful, so moving the lines as in this patch is ok
* wrap the moved line in a `try` / `except` which in the case of an exception does:
  * print `- failed to update cache`
  * set `cache` back to `None`
* instead of the following `else` check the cache again: `if not cache`

That should catch the exception I think you had and gracefully fall back.

mikepurvis:
That's reasonable, I can go that route.

That said, I was definitely stuck in a state which this change as proposed was able to bust me out of— really curious what exactly that was all about, because it wasn't like I was manually mangling anything.

dirk-thomas:
But the proposed route should also get you out, right? For the overhead of rebuilding the cache from scratch (it is still unclear to me why that happened though).

mikepurvis:
As far as root cause, we're definitely still seeing this all the time in our ros_buildfarm instance. Here's the actual exception:

```
$ rosdistro_build_cache index.yaml
Build cache for "indigo"
- trying to use local cache
- use local file "indigo-cache.yaml.gz"
  File "/usr/local/bin/rosdistro_build_cache", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/mikepurvis/rosdistro_py/scripts/rosdistro_build_cache", line 99, in <module>
    main()
  File "/Users/mikepurvis/rosdistro_py/scripts/rosdistro_build_cache", line 80, in main
    caches = generate_distribution_caches(args.index, dist_names=args.dist_names, preclean=args.preclean, ignore_local=args.ignore_local, debug=args.debug)
  File "/Users/mikepurvis/rosdistro_py/src/rosdistro/distribution_cache_generator.py", line 60, in generate_distribution_caches
    cache = generate_distribution_cache(index, dist_name, preclean=preclean, ignore_local=ignore_local, debug=debug)
  File "/Users/mikepurvis/rosdistro_py/src/rosdistro/distribution_cache_generator.py", line 71, in generate_distribution_cache
    dist, cache = _get_cached_distribution(index, dist_name, preclean=preclean, ignore_local=ignore_local)
  File "/Users/mikepurvis/rosdistro_py/src/rosdistro/distribution_cache_generator.py", line 142, in _get_cached_distribution
    cache.update_distribution(rel_file_data)
  File "/Users/mikepurvis/rosdistro_py/src/rosdistro/distribution_cache.py", line 84, in update_distribution
    for repo_name in sorted(new_data['repositories'].keys()):
TypeError: string indices must be integers, not str
```

To `distribution_cache.py` I added the following print to figure out what on earth was going on here:

```
        if len(distribution_file_data) == len(self._distribution_file_data):
            for old_data, new_data in zip(self._distribution_file_data, distribution_file_data):
                print(zip(self._distribution_file_data, distribution_file_data))
                for repo_name in sorted(new_data['repositories'].keys()):
```

The output is:

```
[('repositories', 'repositories'), ('version', 'version'), ('type', 'type'), ('release_platforms', 'release_platforms')]
```

Looks like maybe there's a `keys()` somewhere that's supposed to be `values()`?

mikepurvis:
This is a regression from 0.4.4 to 0.4.5, suspect it came in with #71.

dirk-thomas:
From looking at the exception as well as the code the problem must be in this part: `new_data['repositories']`

Since you run into this frequently can you please try the following snippet instead of the `print`:

````
try:
    new_data_repos = new_data['repositories']
except TypeError:
    print(type(new_data), new_data)
```

Could it be that `repositories` is actually not a dict in your case? E.g. reading something empty from yaml?

mikepurvis:
As expected, given the output above: `<type 'str'> repositories`

The local and remote caches definitely have repositories as a dict:

```
$ head indigo-cache.yaml
distribution_file:
  release_platforms:
    ubuntu: [trusty]
  repositories:
    actionlib:
      doc: {type: git, url: 'https://github.com/ros/actionlib.git', version: indigo-devel}
      release:
        tags: {release: 'release/indigo/{package}/{version}'}
        url: https://github.com/ros-gbp/actionlib-release.git
        version: 1.11.3-0
```

mikepurvis:
Oh dammit, this is REP 143, isn't it. https://github.com/ros/rosdistro/commit/e2229034be6072f888ac46db4c53d0ae6641de9e

I'll update us, but FWIW this is an almost incomprehensible failure mode— it would be much better if it either silently worked, or threw a clear "please update" message.

dirk-thomas:
This should indeed be handled transparently since it shouldn't make a difference for the use - we still allow format 2 if you don't need the new features of format 3.

I have created #73 based on this change and added a small fix to both update functions. Please give it a shot and comment if that works for you too.

PR #73, opened 2016-03-29
Fix updating newer cache when index format 2 is being used

dirk-thomas:
Replaces #72.


mikepurvis:
LGTM.

wjwwood:
lgtm

Just as a note, it looks like this code could be deduplicated a bit since essentially the same change was made in two different files.

PR #75, opened 2016-04-13
Fix rosdistro_freeze_source for REP 143.

mikepurvis:
Does what it says on the tin.

### DIFF
--- a/scripts/rosdistro_freeze_source
+++ b/scripts/rosdistro_freeze_source
@@ -33,6 +33,8 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+from __future__ import print_function
+
 import argparse
 import os.path
 import sys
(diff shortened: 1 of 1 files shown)


scripts/rosdistro_freeze_source
```diff
@@ -75,11 +77,18 @@ def main():
         args.dist_names = sorted(index.distributions.keys())
 
     for dist_name in args.dist_names:
+        dist_url = index.distributions[dist_name]['distribution']
+        if type(dist_url) == list:
+            if len(dist_url) != 1:
+                print("This script only works for distributions with a single distribution.yaml.", file=sys.stderr)
+                continue
+            dist_url = dist_url[0]
```
mikepurvis: I looked at the `Index()` class; I didn't see an obvious way to get this from there in an already version-generic way.

scripts/rosdistro_freeze_source
```diff
@@ -75,11 +77,18 @@ def main():
         args.dist_names = sorted(index.distributions.keys())
 
     for dist_name in args.dist_names:
+        dist_url = index.distributions[dist_name]['distribution']
+        if type(dist_url) == list:
+            if len(dist_url) != 1:
+                print("This script only works for distributions with a single distribution.yaml.", file=sys.stderr)
```
dirk-thomas: Please include the distro name which has been skipped in the error message:

    print("The distribution '%s' has been skipped since this script only works with a single distribution.yaml" % dist_name, file=sys.stderr)

Please use single quotes wherever possible.

scripts/rosdistro_freeze_source
```diff
@@ -75,11 +77,18 @@ def main():
         args.dist_names = sorted(index.distributions.keys())
 
     for dist_name in args.dist_names:
+        dist_url = index.distributions[dist_name]['distribution']
+        if type(dist_url) == list:
```
dirk-thomas: Please use `isinstance(dist_url, list)` instead.

mikepurvis:
Updated.

dirk-thomas:
Thanks!

Issue #66, opened 2016-02-25
rosdistro_build_cache should have an option to cache source repo


mikepurvis:
Via https://github.com/ros-infrastructure/rosdistro/pull/65#issuecomment-187883539:

     It would be better to optionally extend the cache building to explicitly know how to walk source repos. And then with the extended cache the rosinstall_generator could take advantage of the extended cache and fallback to upstream sources.

The biggest issue I see here is that unlike with the GBP release branches, it's harder to tell if a devel branch is stale or not. The best thing you could do would be cache the commit hash, and then use `git ls-remote` to determine if the branch has moved on from that point.

Apart from that, there are two major ways I see this going:

- `rosdistro_build_cache` gets a switch like `--from-source`, and the resulting cache format is unchanged. Has the disadvantage of needing to build a separate cache for source, but the advantage is that other tooling like rosinstall_generator is unchanged.
- The cache format changes, and the source branch package.xml is cached alongside the release one.

Thoughts?

dirk-thomas:
For release repositories the distribution file contains an exact version. Until that changes the cache entry is valid and doesn't have to be rechecked. That is why each cache update only takes a few seconds.

For the source entry the repository state can change anytime and therefore the script must query the state of the repo every time. Each cache update will take quite some time.

Tools like rosinstall_generator should therefore maybe clone the exact hash used for building the cache. Otherwise there will be inconsistencies between the cached information and the cloned repos.

Beside that I don't see any other problem. I would suggest to implement the new option (to build a separate cache) and then gain experience by using it and seeing if anything unforeseen happens.

mikepurvis:
> Tools like rosinstall_generator should therefore maybe clone the exact hash used for building the cache. Otherwise there will be inconsistencies between the cached information and the cloned repos.

Another way that we're looking at to tackle this (and by extension, the business of creating reproducible nightlies) is by a `rosdistro_freeze` script, which creates a single commit to swaps all source/devel branch pointers into hashes, and then tags that frozen branch with a timestamp. From there, we can generate a cache, and create a deterministic "source" workspace from which to generate a build asset.

Would be happy to discuss further how this workflow would go and how we might be able to collaborate on it.

dirk-thomas:
Just looking at the tools we already have and what they do:

* `rosinstall_generator`: generate a `.rosinstall` file from either the release entries, or the upstream repo of the gbp pointing to the exact release tag, or the development branch which is specified in the source entry

[...]

PR #78, opened 2016-04-27
Source manifest caching

mikepurvis:
Addressing #66. ~~This is still missing some pretty major pieces, but I wanted to get it up for a first pass review in its current state, as there are a few bits that could potentially go different ways, and it has already diverged from our earlier discussion, in ways that I will call out below.~~

~~It won't be usable in any meaningful way until there's first-class support for Github, which should be possible to implement using the tree API to locate package.xml files ([eg](https://api.github.com/repos/ros-planning/navigation/git/trees/indigo-devel?recursive=1)), but it's most reasonable to review this in stages, I think.~~

This functionality is complete and ready to merge. Clearpath have been running this in production since it was first proposed in May. This PR has a large diff because it also includes the changes from #79 and #80.

For just the changes here: https://github.com/mikepurvis/rosdistro-1/compare/compact-cache...mikepurvis:source-manifest-caching

----

I made a toy rosdistro to try it on in the present form; you can see the result of generating the cache from this distro here: https://github.com/mikepurvis/rosdistro-minimal/blob/master/kinetic-cache.yaml

### DIFF
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -67,6 +67,9 @@ def parse_args(args=sys.argv[1:]):
     add('--ignore-local', action='store_true', default=False,
         help='Ignore locally available cache'
     )
+    add('--include-source', action='store_true', default=False,
+        help='Also include source branch package XMLs in the cache'
+    )
     return parser.parse_args(args)
 
 
@@ -77,7 +80,8 @@ def main():
         logger.level = logging.DEBUG
 
     try:
-        caches = generate_distribution_caches(args.index, dist_names=args.dist_names, preclean=args.preclean, ignore_local=args.ignore_local, debug=args.debug)
+        caches = generate_distribution_caches(args.index, dist_names=args.dist_names, preclean=args.preclean,
+            ignore_local=args.ignore_local, include_source=args.include_source, debug=args.debug)
     except RuntimeError as e:
         print(str(e), file=sys.stderr)
         sys.exit(1)
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -42,7 +42,7 @@ import sys
 import yaml
 
 from rosdistro import logger
-from rosdistro.distribution_cache_generator import generate_distribution_caches
+from rosdistro.distribution_cache_generator import CacheYamlDumper, generate_distribution_caches
 
 logging.basicConfig(level=logging.INFO)
 
@@ -86,13 +86,19 @@ def main():
         if args.dist_names and dist_name not in args.dist_names:
             continue
         # write the cache
-        data = yaml.dump(cache.get_data())
+        data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper)
+
         with open('%s-cache.yaml' % dist_name, 'w') as f:
             print('- write cache file "%s-cache.yaml"' % dist_name)
             f.write(data)
         with gzip.open('%s-cache.yaml.gz' % dist_name, 'wb') as f:
             print('- write compressed cache file "%s-cache.yaml.gz"' % dist_name)
-            f.write(data.encode('utf-8'))
+            try:
+                # Python 2 gzip module requires encoded data.
+                unicode
+                f.write(data)
+            except NameError:
+                f.write(data.encode('utf-8'))
 
 
 if __name__ == '__main__':
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -92,14 +92,13 @@ def main():
             print('- write cache file "%s-cache.yaml"' % dist_name)
             f.write(data)
         with gzip.open('%s-cache.yaml.gz' % dist_name, 'wb') as f:
+            # On Python 3, we must encode the unicode yaml str prior to gzipping it,
+            # whereas for Python 2, the yaml output is a str which cannot be further
+            # encoded (compared with a unicode object).
+            if sys.version_info[0] >= 3:
+                data = data.encode('utf-8')
             print('- write compressed cache file "%s-cache.yaml.gz"' % dist_name)
-            try:
-                # Python 2 gzip module requires encoded data.
-                unicode
-                f.write(data)
-            except NameError:
-                f.write(data.encode('utf-8'))
-
+            f.write(data)
 
 if __name__ == '__main__':
     main()
--- a/src/rosdistro/distribution.py
+++ b/src/rosdistro/distribution.py
@@ -32,22 +32,30 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from .manifest_provider.bitbucket import bitbucket_manifest_provider
-from .manifest_provider.git import git_manifest_provider
-from .manifest_provider.github import github_manifest_provider
+from .manifest_provider.git import git_manifest_provider, git_source_manifest_provider
+from .manifest_provider.github import github_manifest_provider, github_source_manifest_provider
 
 
 class Distribution(object):
 
     default_manifest_providers = [github_manifest_provider, bitbucket_manifest_provider, git_manifest_provider]
+    default_source_manifest_providers = [github_source_manifest_provider, git_source_manifest_provider]
 
-    def __init__(self, distribution_file, manifest_providers=None):
+    def __init__(self, distribution_file, manifest_providers=None, source_manifest_providers=None):
         self._distribution_file = distribution_file
+
         # Use default
         self._manifest_providers = Distribution.default_manifest_providers
+        self._source_manifest_providers = Distribution.default_source_manifest_providers
+
         # Override default if given
         if manifest_providers is not None:
             self._manifest_providers = manifest_providers
+        if source_manifest_providers is not None:
+            self._source_manifest_providers = source_manifest_providers
+
         self._release_package_xmls = {}
+        self._source_repo_package_xmls = {}
 
     def __getattr__(self, name):
         return getattr(self._distribution_file, name)
@@ -64,11 +72,18 @@ class Distribution(object):
                 return None
             package_xml = None
             for mp in self._manifest_providers:
-                #try:
                 package_xml = mp(self._distribution_file.name, repo, pkg_name)
-                #except:
-                #    pass
                 if package_xml is not None:
                     break
             self._release_package_xmls[pkg_name] = package_xml
         return self._release_package_xmls[pkg_name]
+
+    def get_source_repo_package_xmls(self, repo):
+        """ Expects a SourceRepositorySpecification object. """
+        if not self._source_repo_package_xmls.get(repo.name, None):
+            for mp in self._source_manifest_providers:
+                result = mp(repo)
+                if result is not None:
+                    break
+            self._source_repo_package_xmls[repo.name] = result
+        return self._source_repo_package_xmls[repo.name]
--- a/src/rosdistro/distribution.py
+++ b/src/rosdistro/distribution.py
@@ -78,12 +78,25 @@ class Distribution(object):
             self._release_package_xmls[pkg_name] = package_xml
         return self._release_package_xmls[pkg_name]
 
-    def get_source_repo_package_xmls(self, repo):
-        """ Expects a SourceRepositorySpecification object. """
-        if not self._source_repo_package_xmls.get(repo.name, None):
+    def get_source_package_xml(self, pkg_name):
+        repo_name = self._distribution_file.source_packages[pkg_name].repository_name
+        repo_cache = self.get_source_repo_package_xmls(repo_name)
+        if repo_cache:
+            return repo_cache[pkg_name][1]
+        else:
+            return None
+
+    def get_source_repo_package_xmls(self, repo_name):
+        if repo_name in self._source_repo_package_xmls:
+            return self._source_repo_package_xmls[repo_name]
+        else:
             for mp in self._source_manifest_providers:
-                result = mp(repo)
-                if result is not None:
-                    break
-            self._source_repo_package_xmls[repo.name] = result
-        return self._source_repo_package_xmls[repo.name]
+                repo_cache = mp(self.repositories[repo_name].source_repository)
+                if repo_cache is not None:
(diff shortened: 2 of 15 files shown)


scripts/rosdistro_build_cache
```diff
@@ -67,6 +67,9 @@ def parse_args(args=sys.argv[1:]):
     add('--ignore-local', action='store_true', default=False,
         help='Ignore locally available cache'
     )
+    add('--include-source', action='store_true', default=False,
+        help='Also include source branch package XMLs in the cache'
+    )
```
mikepurvis: For now, I'm implemented this as an option rather than a separate script. I want to understand the file size impact once deduplication is done, and if it's still really bad, then we can revisit making a separate cache file, separate script, etc.

src/rosdistro/distribution.py
```diff
@@ -32,22 +32,30 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from .manifest_provider.bitbucket import bitbucket_manifest_provider
-from .manifest_provider.git import git_manifest_provider
+from .manifest_provider.git import git_manifest_provider, git_source_manifest_provider
 from .manifest_provider.github import github_manifest_provider
```
mikepurvis: There's probably an argument to be made here for moving these to classes with static methods (`GitProvider` with `get_release_package_xml` and `get_source_package_xmls`). I wanted to avoid getting into a big refactoring here, but can do so if there's consensus that that's the way to go.

src/rosdistro/distribution.py
```diff
@@ -32,22 +32,30 @@
...
+
         # Use default
         self._manifest_providers = Distribution.default_manifest_providers
+        self._source_manifest_providers = Distribution.default_source_manifest_providers
+
         # Override default if given
         if manifest_providers is not None:
             self._manifest_providers = manifest_providers
+        if source_manifest_providers is not None:
+            self._source_manifest_providers = source_manifest_providers
```
mikepurvis: This doubling-up would be removed if the providers were classes with static methods (or we passed the module instead of the function).

src/rosdistro/manifest_provider/git.py
```diff
@@ -48,12 +51,49 @@ def git_manifest_provider(_dist_name, repo, pkg_name):
...
+                if 'package.xml' in filenames:
+                    filename = os.path.join(path, 'package.xml')
+                    with open(filename, 'r') as f:
+                        package_xml = f.read()
+
+                    try:
+                        name = parse_package_string(package_xml).name
+                    except InvalidPackage:
+                        raise RuntimeError('Unable to parse package.xml file found in %s' % repo.url)
+                    cache[name] = [ path[len(git_repo_path)+1:], package_xml ]
```
mikepurvis: This format is different than we talked about. It ends up being:

```
source_repo_xmls:
  my_fancy_repo:
    _ref: b858cb282617fb0956d960215c8e84d1ccf909c6
    pkgname1: ['path/to/pkgname1', '<package>....</package>']
    pkgname2: ['path/to/pkgname1', '<package>....</package>']
    pkgname3: ['path/to/pkgname1', '<package>....</package>']
```

Upon reflection, I didn't feel there was value in trying to support multiple hashes in the cache— we're only ever going to cache the "source" version branch anyway. And if we did have multiple in there in a dictionary, then the consuming code no longer has a "default" option; it _has_ to `git ls-remote` to get a hash in order to resolve, which seems like a big step backward to me.

The source repo cache is only ever going to be best-effort, and this format accepts that reality.

mikepurvis:
Okay, this is now fully working for Github. I've generated a cache for Jade, so that we can examine it and compare. The uncompressed file is here:

https://gist.githubusercontent.com/anonymous/3bd2c922bd58e572d928988a925b04bf/raw/ad13562fbe59159e8675924031135e291ccd0b8f/jade-cache.yaml

The size of the current caches are:

```
289975 Apr 28 11:31 jade-cache.yaml.gz
1568414 Apr 28 11:31 jade-cache.yaml
```

The cache with the source package XMLs included is:

```
260410 Apr 28 11:58 jade-cache.yaml.gz
1858121 Apr 28 11:58 jade-cache.yaml
```

So despite that the cache is larger, it actually compresses down smaller (!) than what we currently have. I believe this is attributable to me switching from yaml Dumper to SafeDumper, which forces serialization of objects as strings, rather than representing native Python types. You can see the effect of this in a bunch of places in the current cache, eg:

```
release_package_xmls:
  ackermann_msgs: !!binary |
    PHBhY2thZ2U+CiAgPG5hbWU+YWNrZXJtYW5uX21zZ3M8L25hbWU+CiAgPHZlcnNpb24+MS4wLjE8
    L3ZlcnNpb24+CiAgPGRlc2NyaXB0aW9uPgogICAgUk9TIG1lc3NhZ2VzIGZvciByb2JvdHMgdXNp
    bmcgQWNrZXJtYW5uIHN0ZWVyaW5nLgogIDwvZGVzY3JpcHRpb24+CiAgPG1haW50YWluZXIgZW1h
    aWw9ImphY2sub3F1aW5AZ21haWwuY29tIj5KYWNrIE8nUXVpbjwvbWFpbnRhaW5lcj4KICA8YXV0
    aG9yPkphY2sgTydRdWluPC9hdXRob3I+CiAgPGxpY2Vuc2U+QlNEPC9saWNlbnNlPgoKICA8dXJs
    IHR5cGU9IndlYnNpdGUiPmh0dHA6Ly9yb3Mub3JnL3dpa2kvYWNrZXJtYW5uX21zZ3M8L3VybD4K
    ICA8dXJsIHR5cGU9InJlcG9zaXRvcnkiPmh0dHBzOi8vZ2l0aHViLmNvbS9yb3MtZHJpdmVycy9h
    Y2tlcm1hbm5fbXNncy5naXQ8L3VybD4KICA8dXJsIHR5cGU9ImJ1Z3RyYWNrZXIiPmh0dHBzOi8v
    Z2l0aHViLmNvbS9yb3MtZHJpdmVycy9hY2tlcm1hbm5fbXNncy9pc3N1ZXM8L3VybD4KCiAgPGJ1
    aWxkdG9vbF9kZXBlbmQ+Y2F0a2luPC9idWlsZHRvb2xfZGVwZW5kPgoKICA8YnVpbGRfZGVwZW5k
    Pm1lc3NhZ2VfZ2VuZXJhdGlvbjwvYnVpbGRfZGVwZW5kPgogIDxidWlsZF9kZXBlbmQ+c3RkX21z
    Z3M8L2J1aWxkX2RlcGVuZD4KCiAgPHJ1bl9kZXBlbmQ+bWVzc2FnZV9ydW50aW1lPC9ydW5fZGVw
    ZW5kPgogIDxydW5fZGVwZW5kPnN0ZF9tc2dzPC9ydW5fZGVwZW5kPgogIAogIDxleHBvcnQ+CiAg
    ICA8YXJjaGl0ZWN0dXJlX2luZGVwZW5kZW50Lz4KICA8L2V4cG9ydD4KPC9wYWNrYWdlPgo=
``` 

Whereas with `SafeDumper`, we instead see the more compact, and easier-to-gzip plain text result (including the yaml reference which permits the same string to show up in the source cache section):

```
ackermann_msgs: &id001 "<package>\n  <name>ackermann_msgs</name>\n\
    \  <version>1.0.1</version>\n  <description>\n    ROS messages for robots using\
    \ Ackermann steering.\n  </description>\n  <maintainer email=\"jack.oquin@gmail.com\"\
    >Jack O'Quin</maintainer>\n  <author>Jack O'Quin</author>\n  <license>BSD</license>\n\
    \n  <url type=\"website\">http://ros.org/wiki/ackermann_msgs</url>\n  <url type=\"\
    repository\">https://github.com/ros-drivers/ackermann_msgs.git</url>\n  <url type=\"\
    bugtracker\">https://github.com/ros-drivers/ackermann_msgs/issues</url>\n\n  <buildtool_depend>catkin</buildtool_depend>\n\
    \n  <build_depend>message_generation</build_depend>\n  <build_depend>std_msgs</build_depend>\n\
    \n  <run_depend>message_runtime</run_depend>\n  <run_depend>std_msgs</run_depend>\n\
    \  \n  <export>\n    <architecture_independent/>\n  </export>\n</package>\n"
```

The one caveat to all this is that there's a Github API call required for the github source manifest provider. Calling the API anonymously is limited to 60 hits per hour, so it's highly recommended to generate a token and set the `GITHUB_USER` and `GITHUB_PASSWORD` envvars. Once this is all merged, I definitely feel that the buildfarm should generate source cache entries, but it may make sense to leave the command-line default for it to off, given how slow it would be to do so without the Github API.

Can't merge until I have a story for refreshing the cache, but I feel this is meaningful progress and would interested in comments.

src/rosdistro/distribution_cache.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, name, data=None, distribution_file_data=None):
         self._distribution_file_data = data['distribution_file'] if data else distribution_file_data
         self.distribution_file = create_distribution_file(name, self._distribution_file_data)
         self.release_package_xmls = data['release_package_xmls'] if data else {}
+        self.source_repo_package_xmls = data['source_repo_package_xmls'] if data and 'source_repo_package_xmls' in data else {}
```
dirk-thomas: Might this be a problem if the key exists but is empty? Then the variable will not be a dict as later expected.

src/rosdistro/manifest_provider/git.py
```diff
@@ -47,102 +49,76 @@
...
-        _git_client_version = result['output'].split()[-1]
-    return LooseVersion(_git_client_version) >= LooseVersion(version)
+def git_source_manifest_provider(repo):
+    try:
+        with _temp_git_clone(repo.url, repo.version) as git_repo_path:
+            # Include the git hash in our cache dictionary.
+            result = Git(git_repo_path).command('rev-parse', 'HEAD')
+            cache = { '_ref': result['output'] }
+
+            for path, dirnames, filenames in os.walk(git_repo_path):
```
dirk-thomas: This loop is reimplementing `catkin_pkg` functionality and e.g. doesn't honor `CATKIN_IGNORE` files, doesn't check for duplicate package names.

Can this reuse [catkin_pkg.packages.find_package_paths](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/packages.py#L41) or [catkin_pkg.packages.find_packages](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/packages.py#L71)?

src/rosdistro/manifest_provider/git.py
dirk-thomas: Sounds like a very good idea to me.

src/rosdistro/manifest_provider/git.py
```diff
@@ -47,102 +49,76 @@
...
+        raise RuntimeError('Unable to fetch source package.xml files: %s' % e)
+
+    return cache
 
 
-def _get_package_xml(url, tag):
+@contextmanager
+def _temp_git_clone(url, ref):
     base = tempfile.mkdtemp('rosdistro')
+    git = Git(base)
```
dirk-thomas: Maybe pass `base` explicitly as a keyword argument. That makes the following code using `.` more clear:

```
git  = Git(cwd=base)
```

src/rosdistro/manifest_provider/git.py
mikepurvis: Agree, will change.

src/rosdistro/distribution_cache.py
mikepurvis: I can put in a fallback for that, but I don't think there's any codepath which would cause the `source_repo_package_xmls` key to be present but not equal to a dictionary, so I'd consider such a distribution.yaml to be invalid.

src/rosdistro/manifest_provider/git.py
mikepurvis: Done.

src/rosdistro/manifest_provider/git.py
mikepurvis: Resolved in the earlier PR. ;)

mikepurvis:
Rebased this one again too, and ran the following sanity checks:

```
git clone https://github.com/ros/rosdistro
cd rosdistro
rosdistro_build_cache --debug index.yaml kinetic

export GITHUB_USER=xxx; export GITHUB_PASSWORD=yyy  # Optional, makes the github provider much faster since it can use the API instead of full clones.
rosdistro_build_cache --debug --include-source index.yaml kinetic
```

The result of the second invocation is viewable here: https://gist.github.com/mikepurvis/ff2b5787a7741c07f560ee370ac9e363

Note that the behaviour to correctly detect changed repos and drop the associated cache entries when _updating_ is added in #84, here: https://github.com/ros-infrastructure/rosdistro/pull/84/files#diff-f5dd4132cfc7d911ffc46fbed75a04fcR118

src/rosdistro/manifest_provider/cache.py
```diff
@@ -42,19 +79,59 @@ def __init__(self, distribution_cache, manifest_providers=None):
...
+
+
+class CachedSourceManifestProvider(object):
+
+    def __init__(self, distribution_cache, source_manifest_providers=None):
+        self._distribution_cache = distribution_cache
+        self._source_manifest_providers = source_manifest_providers
+
+    def __call__(self, repo):
+        assert repo.url and repo.version
```
Rayman: SVN repositories don't have a version. Does this work or is this checking something else?

dirk-thomas: @mikepurvis Can you comment on this question?

mikepurvis: The assert will fail and no cache will be created (the same handling as for hg repos).

mikepurvis: I spoke too soon, and had inadequately tested for this case. I'm using a test rosdistro now which includes hg and non-version svn entries. Commit forthcoming.

Rayman: Thanks :+1:

PR #79, opened 2016-04-27
Refactor manifest provider utility functions

mikepurvis:
Some more stuff I need for #78 that will be helpful to merge in advance:

- Created new `rosdistro.vcs` module with `Git` class having `command` and `version_gte` methods, to take the place of the loose stuff that used to be in `rosdistro.manifest_providers.git`.
- Moved `get_release_tag` from `rosdistro.manifest_providers` to a member function of `ReleaseRepositorySpecification`.
- Dropped `check_remote_tag_exists` from `rosdistro.manifest_providers.git`, in favour of a property on `RepositorySpecification` that returns (with caching) remote references and remote tags.
- Added some tests to cover the basic functionality of the manifest providers.

@dirk-thomas

### DIFF
--- a/src/rosdistro/manifest_provider/__init__.py
+++ b/src/rosdistro/manifest_provider/__init__.py
@@ -32,6 +32,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 
+# This function remains here as a compatibility shim because there
+# is code in rosinstall_generator which uses it in this location.
 def get_release_tag(repo, pkg_name):
-    """ Compatibility shim. """
     return repo.get_release_tag(pkg_name)
--- a/src/rosdistro/manifest_provider/__init__.py
+++ b/src/rosdistro/manifest_provider/__init__.py
@@ -33,13 +33,5 @@
 
 
 def get_release_tag(repo, pkg_name):
-    data = {
-        'package': pkg_name
-    }
-    if repo.version is not None:
-        data['version'] = repo.version
-        data['upstream_version'] = repo.version.split('-')[0]
-    release_tag = repo.tags['release']
-    for k, v in data.items():
-        release_tag = release_tag.replace('{%s}' % k, v)
-    return release_tag
+    """ Compatibility shim. """
+    return repo.get_release_tag(pkg_name)
--- a/src/rosdistro/manifest_provider/bitbucket.py
+++ b/src/rosdistro/manifest_provider/bitbucket.py
@@ -43,8 +43,6 @@ import os
 import re
 
 from rosdistro import logger
-from rosdistro.manifest_provider import get_release_tag
-from rosdistro.manifest_provider.git import check_remote_tag_exists
 
 # This Bitbucket provider can optionally send basic auth to fetch from
 # private repositories if you supply the environment variables
@@ -56,37 +54,29 @@ from rosdistro.manifest_provider.git import check_remote_tag_exists
 BITBUCKET_USER = os.getenv('BITBUCKET_USER', None)
 BITBUCKET_PASSWORD = os.getenv('BITBUCKET_PASSWORD', None)
 
-# Three capture groups are server, org, repo_name. URLs are expected to be well-behaved
-# rosdistro-style URLs either in https or ssh format.
-GIT_REGEX = re.compile('(?:https?:\/\/|ssh:\/\/|git@)([\w.-]+)[:/]([\w-]+)\/([\w-]*)\.git$')
-
 
 def bitbucket_manifest_provider(_dist_name, repo, pkg_name):
     assert repo.version
+    server, path = repo.get_url_parts()
 
-    match = GIT_REGEX.match(repo.url)
-    if not match:
-        raise RuntimeError('Unexpected git URL format: %s' % repo.url)
-
-    server, org, repo_name = match.groups()
     if server != 'bitbucket.org':
         logger.debug('Skip non-bitbucket url "%s"' % repo.url)
         raise RuntimeError('Cannot handle non bitbucket url.')
 
-    release_tag = get_release_tag(repo, pkg_name)
+    release_tag = repo.get_release_tag(pkg_name)
 
-    if not check_remote_tag_exists(repo.url, release_tag):
+    if release_tag not in repo.remote_tags:
         raise RuntimeError('specified tag "%s" is not a git tag' % release_tag)
 
-    url = 'https://%s/%s/%s/raw/%s/package.xml' % (server, org, repo_name, release_tag)
+    url = 'https://bitbucket.org/%s/raw/%s/package.xml' % (path, release_tag)
     try:
         logger.debug('Load package.xml file from url "%s"' % url)
         req = Request(url)
         if BITBUCKET_USER and BITBUCKET_PASSWORD:
             logger.debug('- using http basic auth from supplied environment variables.')
-            authheader = 'Basic %s' % base64.encodestring('%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
+            authheader = 'Basic %s' % base64.b64encode('%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
             req.add_header('Authorization', authheader)
-        package_xml = urlopen(req).read()
+        package_xml = urlopen(req).read().decode('utf-8')
         return package_xml
     except URLError as e:
         logger.debug('- failed (%s)' % e)
--- a/src/rosdistro/manifest_provider/bitbucket.py
+++ b/src/rosdistro/manifest_provider/bitbucket.py
@@ -65,7 +65,7 @@ def bitbucket_manifest_provider(_dist_name, repo, pkg_name):
 
     release_tag = repo.get_release_tag(pkg_name)
 
-    if release_tag not in repo.remote_tags:
+    if not repo.has_remote_tag(release_tag):
         raise RuntimeError('specified tag "%s" is not a git tag' % release_tag)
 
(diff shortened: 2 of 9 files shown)


test/test_index.py
```diff
@@ -55,7 +55,7 @@ def test_get_index_from_http_with_query_parameters():
     else:
         proc = subprocess.Popen([sys.executable, '-m', 'http.server', '9876'],
                                 cwd=FILES_DIR)
-    time.sleep(0.1)
+    time.sleep(0.5)
```
mikepurvis: Upping this time made the test less flaky for me.

test/test_repository_specification.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_repository_specification():
+    data = { 'type': 'git', 'url': 'https://github.com/ros/catkin.git' }
+    r = RepositorySpecification("test", data)
+    assert r.get_data() == data
+    assert r.version == None
+    assert r.get_url_parts() == ('github.com', 'ros/catkin')
+
+    r.url = 'http://github.com/ros/catkin'
+    assert r.get_url_parts() == ('github.com', 'ros/catkin')
```
mikepurvis: I would have made this only accept the `.git` suffix, but there are URLs in `ros/rosdistro` that haven't got it.

mikepurvis:
One other bit of miscellaneous functionality that it would be nice to move into RepositorySpecification is getting the remotes tags and branches, especially if it could be a method that internally cached results between invocations, so that multiple manifest providers could use the same information about whether tags exist and what ref the devel branch is associated with (for example, an additional shortcut is possible for the source manifest cacher if the hash of the tag that corresponds to the latest version is the same as the hash of the devel branch).

The main cost of this is that then we'll have a different place capable of invoking git commands, so it might make sense to also relocate {{_find_executable}} and {{_run_command}} to somewhere more central (possibly collapsing them to a single {{execute_git}} function that transparently handles finding the executable on first invocation.

test/test_manifest_providers.py
```diff
@@ -0,0 +1,55 @@
...
+        'version': '0.5.7-1'
+    })
+
+
+def _rospeex_repo():
+    return ReleaseRepositorySpecification('rospeex', {
+        'packages': [ 'rospeex', 'rospeex_msgs' ],
+        'tags': { 'release': 'release/indigo/{package}/{version}' },
+        'url': 'https://bitbucket.org/rospeex/rospeex-release.git',
+        'version': '2.14.7-0'
```
mikepurvis: I picked this out of the indigo rosdistro. Is there a better one to use?

mikepurvis:
@dirk-thomas This is ready for review now, along with #80.

src/rosdistro/manifest_provider/__init__.py
```diff
@@ -33,13 +33,5 @@
...
-    }
-    if repo.version is not None:
-        data['version'] = repo.version
-        data['upstream_version'] = repo.version.split('-')[0]
-    release_tag = repo.tags['release']
-    for k, v in data.items():
-        release_tag = release_tag.replace('{%s}' % k, v)
-    return release_tag
+    """ Compatibility shim. """
+    return repo.get_release_tag(pkg_name)
```
mikepurvis: Shim is necessary because the current version of rosinstall_generator uses this function.

src/rosdistro/manifest_provider/bitbucket.py
```diff
@@ -56,37 +54,29 @@
...
     try:
         logger.debug('Load package.xml file from url "%s"' % url)
         req = Request(url)
         if BITBUCKET_USER and BITBUCKET_PASSWORD:
             logger.debug('- using http basic auth from supplied environment variables.')
-            authheader = 'Basic %s' % base64.encodestring('%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
+            authheader = 'Basic %s' % base64.b64encode('%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
             req.add_header('Authorization', authheader)
-        package_xml = urlopen(req).read()
+        package_xml = urlopen(req).read().decode('utf-8')
```
mikepurvis: Without this change, the new unit tests fail under Python 3 with the error `Type str doesn't support the buffer API`. Full conversion of the cache to native unicode is in #80.

src/rosdistro/manifest_provider/github.py
```diff
@@ -39,32 +39,24 @@
...
-    url += '/%s/package.xml' % release_tag
-    if url.startswith('git://'):
-        url = 'https://' + url[6:]
-    if url.startswith('https://'):
-        url = 'https://raw.' + url[8:]
+    url = 'https://raw.githubusercontent.com/%s/%s/package.xml' % (path, release_tag)
     try:
         logger.debug('Load package.xml file from url "%s"' % url)
-        package_xml = urlopen(url).read()
+        package_xml = urlopen(url).read().decode('utf-8')
```
mikepurvis: See note: https://github.com/ros-infrastructure/rosdistro/pull/79/files#r72559554

src/rosdistro/manifest_provider/__init__.py
```diff
@@ -33,13 +33,5 @@
...
-        'package': pkg_name
-    }
-    if repo.version is not None:
-        data['version'] = repo.version
-        data['upstream_version'] = repo.version.split('-')[0]
-    release_tag = repo.tags['release']
-    for k, v in data.items():
-        release_tag = release_tag.replace('{%s}' % k, v)
-    return release_tag
+    """ Compatibility shim. """
```
dirk-thomas: The docblock is not really helpful. I would suggest remove it again and adding a plain comment above the function mentioning that it is being used by `rosinstall_generator`.

src/rosdistro/manifest_provider/github.py
```diff
@@ -39,32 +39,24 @@
...
+    server, path = repo.get_url_parts()
+    if server != 'github.com':
         logger.debug('Skip non-github url "%s"' % repo.url)
         raise RuntimeError('can not handle non github urls')
 
-    release_tag = get_release_tag(repo, pkg_name)
+    release_tag = repo.get_release_tag(pkg_name)
 
-    if not check_remote_tag_exists(repo.url, release_tag):
+    if release_tag not in repo.remote_tags:
```
dirk-thomas: By changing the way of the condition ("does tag x exist" vs. "x in list-of-tags") this limits future implementations since the API now always needs to get all remote tags. Can we change this to be a method on `RepositorySpecification` but with the old semantic (`has_remote_tag(tag_name)`)?

src/rosdistro/vcs.py
```diff
@@ -0,0 +1,80 @@
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2016, Clearpath Robotics
```
dirk-thomas: Since some part of this code was refactored from another file I think it should keep the previous copyright statement (in addition to Clearpath).

test/test_repository_specification.py
dirk-thomas: Can you point to the entries which don't? I thought this would be checked by the tests in the repo and therefore by Travis.

test/test_manifest_providers.py
dirk-thomas: Should be fine as-is.

src/rosdistro/manifest_provider/github.py
mikepurvis: I suppose I haven't really made good on it, but the thought was that there would be other places where it would be valuable to be able to iterate on the remote tags, for example:

https://github.com/ros-infrastructure/rosdistro/blob/856aa03ce8b9517c18e5a0787d8f5f2420ecf7eb/src/rosdistro/freeze_source.py#L90

FWIW, this API would allow testing membership without getting the full list, albeit with a little more magic behind the scenes: https://docs.python.org/2/reference/datamodel.html#object.__contains__

src/rosdistro/vcs.py
mikepurvis: Good point, will change.

src/rosdistro/manifest_provider/github.py
mikepurvis: As of dffc5c566aa720b6044d011bb54629274aa37b3e, I've addressed everything here except this. Would you still like me to switch to (or add) a `has_remote_tag` method?

dirk-thomas: Yes, I think it would be good to add a `has_remote_tag` method and use it here. The current repo impl. can still continue to get all tags and implement that method by only considering the already fetch data.

`remote_tags` would just not be public API which needs to be implemented by alternative impl. It allows an alternative implementation to check the requested tag on demand.

src/rosdistro/manifest_provider/git.py
```diff
@@ -47,51 +45,39 @@
...
-            cmd = [_git_client_executable, 'clone', url, base, '--depth', '1', '--branch', tag]
-            result = _run_command(cmd, base)
+            result = git.command('clone', url, base, '--depth', '1', '--branch', tag)
             if result['returncode'] != 0:
                 raise RuntimeError('Could not clone repository "%s" at tag "%s"' % (url, tag))
         else:
             # Old git doesn't support cloning a tag directly, so check it out after a full clone.
-            cmd = [_git_client_executable, 'clone', url, base]
-            result = _run_command(cmd, base)
+            git = Git(base)
```
dirk-thomas: Why create a new instance? Can't the following code use the instance created in line 58?

mikepurvis:
Feedback addresses, and freshly rebased on master.

PR #80, opened 2016-04-30
Add XML squashing and specific YAML dumper.

mikepurvis:
Another piece of the #78 puzzle. This reduces the current size of the gzipped indigo cache from 566k to 293k, a savings of almost exactly 50%, by making the following changes:

- Strip comments and whitespace from the cached package XMLs.
- Supply a yaml dumper config that doesn't insert line continuations everywhere.
- Dump native unicode rather than ascii.

I've verified that the new format works with rosinstall_generator, including with packages whose XML includes unicode characters, eg:

```
# Change the cache point in index.yaml, then:
$ export ROSDISTRO_INDEX_URL=file:///home/mpurvis/rosdistro/index.yaml
$ rosinstall_generator --rosdistro indigo aruco
- git:
    local-name: aruco_ros/aruco
    uri: https://github.com/bmagyar/aruco_ros-release.git
    version: release/indigo/aruco/0.1.0-0
```

@dirk-thomas I believe this is good to review/merge.

----

This is a follow-on to #79. Standalone diff: https://github.com/mikepurvis/rosdistro-1/compare/refactor-git-url-regex...mikepurvis:compact-cache

### DIFF
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -42,7 +42,7 @@ import sys
 import yaml
 
 from rosdistro import logger
-from rosdistro.distribution_cache_generator import generate_distribution_caches
+from rosdistro.distribution_cache_generator import CacheYamlDumper, generate_distribution_caches
 
 logging.basicConfig(level=logging.INFO)
 
@@ -86,13 +86,19 @@ def main():
         if args.dist_names and dist_name not in args.dist_names:
             continue
         # write the cache
-        data = yaml.dump(cache.get_data())
+        data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper)
+
         with open('%s-cache.yaml' % dist_name, 'w') as f:
             print('- write cache file "%s-cache.yaml"' % dist_name)
             f.write(data)
         with gzip.open('%s-cache.yaml.gz' % dist_name, 'wb') as f:
             print('- write compressed cache file "%s-cache.yaml.gz"' % dist_name)
-            f.write(data.encode('utf-8'))
+            try:
+                # Python 2 gzip module requires encoded data.
+                unicode
+                f.write(data)
+            except NameError:
+                f.write(data.encode('utf-8'))
 
 
 if __name__ == '__main__':
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -92,14 +92,13 @@ def main():
             print('- write cache file "%s-cache.yaml"' % dist_name)
             f.write(data)
         with gzip.open('%s-cache.yaml.gz' % dist_name, 'wb') as f:
+            # On Python 3, we must encode the unicode yaml str prior to gzipping it,
+            # whereas for Python 2, the yaml output is a str which cannot be further
+            # encoded (compared with a unicode object).
+            if sys.version_info[0] >= 3:
+                data = data.encode('utf-8')
             print('- write compressed cache file "%s-cache.yaml.gz"' % dist_name)
-            try:
-                # Python 2 gzip module requires encoded data.
-                unicode
-                f.write(data)
-            except NameError:
-                f.write(data.encode('utf-8'))
-
+            f.write(data)
 
 if __name__ == '__main__':
     main()
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -138,7 +138,7 @@ class CacheYamlDumper(yaml.SafeDumper):
     def represent_mapping(self, tag, mapping, flow_style=False):
         """ Gives compact representation for the distribution_file section, while allowing the package
             XML cache sections room to breathe."""
-        if any([ x in mapping for x in ('source', 'release', 'doc')]):
+        if any([x in mapping for x in ('source', 'release', 'doc')]):
             flow_style = True
         return yaml.SafeDumper.represent_mapping(self, tag, mapping, flow_style)
 
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -113,6 +113,31 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
     return cache
 
 
+class CacheYamlDumper(yaml.SafeDumper):
+    def __init__(self, *args, **kwargs):
+        """ Allow long lines and direct unicode representation. This avoids writing escape sequences,
+            line continuations, and other noise into the cache file. """
+        kwargs['width'] = 10000
+        kwargs['allow_unicode'] = True
+        super(CacheYamlDumper, self).__init__(*args, **kwargs)
+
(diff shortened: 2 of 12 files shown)


mikepurvis:
Example result, from Jade: https://gist.githubusercontent.com/anonymous/c46ad9f5023fbf108a4f3b1dc7550ab7/raw/5e9d301e1cfa335c9f183d53e653ee4fcfe88442/jade-cache.yaml

You can see the unicode representation working if you open the link and search for `aruco:`.

scripts/rosdistro_build_cache
```diff
@@ -86,13 +86,19 @@ def main():
...
             f.write(data)
         with gzip.open('%s-cache.yaml.gz' % dist_name, 'wb') as f:
             print('- write compressed cache file "%s-cache.yaml.gz"' % dist_name)
-            f.write(data.encode('utf-8'))
+            try:
+                # Python 2 gzip module requires encoded data.
+                unicode
+                f.write(data)
+            except NameError:
+                f.write(data.encode('utf-8'))
```
dirk-thomas: Wouldn't this look "nicer" and have less repetition?

```
try:
    data = data.encode('utf-8')
except AttributeError:
    pass
f.write(data)
```

src/rosdistro/distribution_cache_generator.py
```diff
@@ -113,6 +113,31 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
     return cache
 
 
+class CacheYamlDumper(yaml.SafeDumper):
+    def __init__(self, *args, **kwargs):
+        """ Allow long lines and direct unicode representation. This avoids writing escape sequences,
+            line continuations, and other noise into the cache file. """
```
dirk-thomas: I now the existing code base is not really "clean" but it would be nice for new docblocks to follow PEP257.

src/rosdistro/manifest_provider/cache.py
```diff
@@ -32,6 +32,35 @@
...
+        return node
+    try:
+        # Python 2. The minidom module parses as ascii, so we have to pre-encode.
+        if isinstance(xml_string, unicode):
+            xml_string = xml_string.encode('utf-8')
+        # Returns an encoded str.
+        return _squash(minidom.parseString(xml_string)).toxml('utf-8')
+    except NameError:
+        # Python 3 returns a unicode str.
+        return _squash(minidom.parseString(xml_string)).toxml()
```
dirk-thomas: I always try to avoid as much duplicate logic as possible. The chances that one of the lines is changed in the future is high. Maybe something like this:

```
try:
    xml_string = xml_string.encode('utf-8')
except AttributeError:
    pass
xml_node = _squash(minidom.parseString(xml_string))
try:
    unicode
    return xml_node.toxml('utf-8')
except NameError:
    return xml_node.toxml()  # return unicode
```

scripts/rosdistro_build_cache
mikepurvis: The `encode

dirk-thomas: What about the following code then. I would think it is easier to read and has no duplication:

```
if sys.version_info[0] >= 3:
    data = data.encode('utf-8')
f.write(data)
```

I am just surprised that the data doesn't need to be encoded in Python 2...

mikepurvis: I can confirm that when always encoding, Python 2 craps out with:

```
  File "/usr/local/bin/rosdistro_build_cache", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/mikepurvis/rosdistro_py/scripts/rosdistro_build_cache", line 99, in <module>
    main()
  File "/Users/mikepurvis/rosdistro_py/scripts/rosdistro_build_cache", line 96, in main
    f.write(data.encode('utf-8'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 132821: ordinal not in range(128)
```

Your final suggestion is reasonable; I shall change it to that.

dirk-thomas: The original code was calling `encode` unconditionally. The error message sounds more like you are using it in an environment which doesn't have a `utf-8` locale (hence the decode error with the `ascii` codec).

mikepurvis: I would note here that the "real" solution would be for `data` to always be a Python 3 `str` or a Python 2 `unicode`, so that encoding it is always a valid operation. However, it seems like pyyaml on Python 2 has no such option, and it is only achievable by hacks.

mikepurvis: Changed, with an added comment explaining what is going on.

src/rosdistro/distribution_cache_generator.py
mikepurvis: Fixed, and moved the docblock to the class rather than constructor.

src/rosdistro/manifest_provider/cache.py
mikepurvis: Done, although AttributeError on encode is not adequate to detect Py2, since the `str` type in Python 2 _has_ an `encode` method which fails in other ways. I believe what I have now works and is in line with your suggestion here.

mikepurvis:
Feedback addressed, rebased onto the changes from #79.

src/rosdistro/manifest_provider/git.py
```diff
@@ -47,102 +49,44 @@
...
+@contextmanager
+def _temp_git_clone(url, ref):
     base = tempfile.mkdtemp('rosdistro')
+    git = Git(base)
     try:
-        assert _git_client_executable is not None, "'git' not found"
-        if _git_version_gte('1.8.0'):
-            # Directly clone the required tag with least amount of additional history. This behaviour
-            # has been available since git 1.8.0:
+        git = Git(cwd=base)
```
dirk-thomas: Isn't this instance redundant again?

dirk-thomas:
Thank you for iterating on this. I made a small PR addressing some style stuff (mikepurvis/rosdistro-1#1) in order to not make too may nitpick comments :wink:

The only "functional" change I see is that the cache now returns sanitized (so modified) xml. I am not sure if anyone will wonder why the xml is different when coming from the cache. What is your experience with the cache size with the new optimizations (referencing other strings, removing extra whitespace)? How much of a difference does it make, especially when being gzipped?

This looks good to me. @ros/ros_team Any comments?

mikepurvis:
Thanks for the review and discussion, much appreciated. I'm looking forward to having this all merged and available to the community.

Re compacted XML: Agree that this is change. Given that the 99% use case is dependency resolution, IMO it is worth it— noting as well that a user who desires the _original_ XML can trivially use the manifest provider functions to acquire it themselves.

mikepurvis:
ping :)

tfoote:
If we can save 50% by sanitizing the xml that seems like a good enough trade off.

dirk-thomas:
See #111 for a follow up fix of the `sanitize` function.

PR #83, opened 2016-07-28
Better matching for tag/branch freezing.

mikepurvis:
The previous logic would catch on a branch name like `fixup-to-1.1.0`, where this insists on a match.

### DIFF
--- a/src/rosdistro/freeze_source.py
+++ b/src/rosdistro/freeze_source.py
@@ -91,10 +91,10 @@ def _worker(work_queue):
             ls_remote_lines = subprocess.check_output(cmd).splitlines()
             for line in ls_remote_lines:
                 hash, ref = line.split('\t', 1)
-                if ref == 'refs/tags/%s' % freeze_version and freeze_to_tag:
+                if freeze_to_tag and ref == 'refs/tags/%s' % freeze_version:
                     source_repo.version = ref.split('refs/tags/')[1]
                     break
-                elif ref == 'refs/heads/%s' % freeze_version or ref == 'refs/tags/%s' % freeze_version:
+                elif ref in ('refs/heads/%s' % freeze_version, 'refs/tags/%s' % freeze_version):
                     source_repo.version = hash
                     break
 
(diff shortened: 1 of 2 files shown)


src/rosdistro/freeze_source.py
```diff
@@ -91,11 +91,11 @@ def _worker(work_queue):
             ls_remote_lines = subprocess.check_output(cmd).splitlines()
             for line in ls_remote_lines:
                 hash, ref = line.split('\t', 1)
-                if ref.endswith(freeze_version):
-                    if freeze_to_tag and ref.startswith('refs/tags/'):
-                        source_repo.version = ref.split('refs/tags/')[1]
-                    else:
-                        source_repo.version = hash
+                if ref == 'refs/tags/%s' % freeze_version and freeze_to_tag:
```
dirk-thomas: I think swapping the order makes the condition a bit easier to read, first the important flag then the "messy" detail:

```
if freeze_to_tag and ref == 'refs/tags/%s' % freeze_version:
```

src/rosdistro/freeze_source.py
```diff
@@ -91,11 +91,11 @@ def _worker(work_queue):
...
+                    source_repo.version = ref.split('refs/tags/')[1]
+                    break
+                elif ref == 'refs/heads/%s' % freeze_version or ref == 'refs/tags/%s' % freeze_version:
```
dirk-thomas: The following might be easier to read:

```
elif ref in ['refs/heads/%s' % freeze_version, 'refs/tags/%s' % freeze_version]:
```

dirk-thomas:
Could you update the patch based on the minor comments? Thank you.

mikepurvis:
Done.

dirk-thomas:
Thank you for adding all these new features to `rosdistro` :clap:

mikepurvis:
I appreciate your patience over a long review cycle.

Looking forward to being able to install rosdistro and rosinstall_generator in our CI environment from debs rather than pip SSH urls. :)

dirk-thomas:
With the current state can you start using the current master branch? Or are you already using that state of the package (via pip ssh url)?

mikepurvis:
Yes, I will switch us over to the current master branch to begin validating prior to your next release.

dirk-thomas:
Great, thank you.

mikepurvis:
We're now running our builds from ros-infrastructure/rosdistro#master and mikepurvis/rosinstall_generator#use-source-walker; cache updating and bundle builds are going through successfully.

I'll try to get a PR together for you in the next couple days for the age-controlled cache updating.

PR #84, opened 2016-07-28
SourceDependencyWalker, and cache updating

mikepurvis:
This is the final piece of the source manifest caching puzzle— the bit which exposes an easy tie-in for rosinstall_generator and similar tooling. This is a follow-on to the following previous PRs, which have all now been rebased for the following review/merge order:

- #79 <- Generalizing git command line functionality, added tests for manifest providers.
- #80 <- Cache size reduction.
- #78 <- Implement source manifest caching.

We've actually been running all of this in prod since early June, so what you're seeing here has some miles under it (barring any horrible mistakes in the rebasing and cherry-picking necessary to prepare these four updated sequential PRs).

For the diff of just this one: https://github.com/mikepurvis/rosdistro-1/compare/source-manifest-caching...mikepurvis:source-dep-resolution-pr

@dirk-thomas @tfoote

### DIFF
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -71,7 +71,8 @@ def generate_distribution_caches(index, dist_names=None, preclean=False,
 
 def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=False,
                                 include_source=False, debug=False):
-    dist, cache = _get_cached_distribution(index, dist_name, preclean=preclean, ignore_local=ignore_local)
+    dist, cache = _get_cached_distribution(index, dist_name, preclean=preclean, ignore_local=ignore_local,
+                                           include_source=include_source)
 
     print('- fetch missing release manifests')
     errors = []
@@ -114,8 +115,7 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
         print('- fetch source repository manifests')
         for repo_name in sorted(dist.repositories.keys()):
             if dist.repositories[repo_name].source_repository:
-                source_repo_cache = dist.get_source_repo_package_xmls(
-                    dist.repositories[repo_name].source_repository)
+                source_repo_cache = dist.get_source_repo_package_xmls(repo_name)
                 if debug:
                     print('  - fetch "%s"' % repo_name)
                 else:
@@ -165,7 +165,7 @@ class CacheYamlDumper(yaml.SafeDumper):
         return yaml.SafeDumper.represent_mapping(self, tag, mapping, flow_style)
 
 
-def _get_cached_distribution(index, dist_name, preclean=False, ignore_local=False):
+def _get_cached_distribution(index, dist_name, preclean=False, ignore_local=False, include_source=False):
     print('Build cache for "%s"' % dist_name)
     cache = None
     try:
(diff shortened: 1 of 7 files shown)


src/rosdistro/distribution_cache_generator.py
```diff
@@ -193,7 +193,11 @@ def _get_cached_distribution(index, dist_name, preclean=False, ignore_local=Fals
         # since format 2 of the index file might contain a single value rather then a list
         if not isinstance(rel_file_data, list):
             rel_file_data = [rel_file_data]
-        # update cache with current distribution file
+        # if we're not including the source portion of the cache, strip it out of the existing cache
+        # in order to skip the potentially lengthy cache invalidation process.
+        if not include_source:
+            cache.source_repo_package_xmls = {}
+        # update cache with current distribution file, which filters existing cache by validity.
```
mikepurvis: This is the one piece I'm not certain about. We have a small enough rosdistro internally that having to constantly ls-remote all the sources to keep the cache fresh is a fine tradeoff. However, for the buildfarm, you would probably want a `--retain-cache` option, which doesn't freshen the source manifest cache, but doesn't drop it either.

In any case, since `--include-source` is an opt-in right now, I think this is fine to merge as-is. At such time as you're wanting to enable source caching by default in ros_buildfarm, we could look at an option to allow only freshening the source cache once a day or whatever.

dirk-thomas:
I am sorry that I haven't reviewed any of the PRs yet, but I currently have to focus on different projects. I will try my best to get to them in the near future.

mikepurvis:
@dirk-thomas Anything I can do here to help increase clarity and/or streamline the review process?

dirk-thomas:
The other three PRs (#79, #80, #78) looks good to me. One has a questions pending. Does this one needs to be rebased?

mikepurvis:
Do you have thoughts on the business of an option to retain but not update the source cache?

Or alternatively, a field in the cache file itself for "time of last update", and then some way of specifying a desired freshening interval.

dirk-thomas:
Adding the time when the source part of the cache was last updated sounds like a good idea.

The new option `--include-source` could get an optional "minimum age" argument. If not passed it will update, if specified it will only do so if the cache if older then the specified age.

We can either distinguish between two options (one to keep the existing source cache, one for the updating condition) or we could use one and you would need to pass e.g. `99years` as the update condition to keep an existing source cache without updating it. I can be convinced to do either or.

mikepurvis:
Okay, have rebased at least, but this will have to wait for me to implement the min-age scheme discussed above (or that could come as a follow-up, whatever).

Either way, I believe the previous three are good to go in, though if you use the github squash merge, I may need to actively rebase each one as you do it, not sure about that.

dirk-thomas:
I just merged the other three PRs. Would you like this to be merged or keep it on hold until you have added the min-age scheme?

mikepurvis:
Best to get it in, I think. Adding min-age won't interfere with the existing `--include-source` argument, I don't think, and having this in allows the rosinstall_generator PR to go up and be reviewed.

src/rosdistro/distribution_cache_generator.py
dirk-thomas: I agree with you statement. The ROS build farm will likely not enable `--include-source` until we have a `--retain-source-cache` option.

src/rosdistro/manifest_provider/git.py
```diff
@@ -87,9 +85,9 @@ def _temp_git_clone(url, ref):
     base = tempfile.mkdtemp('rosdistro')
     git = Git(cwd=base)
     try:
-        if git.version_gte('1.8.0') and not _ref_is_hash(ref):
-            # Directly clone the required ref with least amount of additional history.
-            # Available since git 1.8.0, but only works for tags and branches, not hashes:
+        if git.version_gte('1.8.0') and not ref_is_hash(ref):
+            # Directly clone the required ref with least amount of additional history. This behaviour
+            # has been available since git 1.8.0, but only works for tags and branches, not hashes:
```
dirk-thomas: Can you please undo the comment change? I think this was something we iterated on in the other PRs and was intentionally changed to the shorter left side.

mikepurvis: Done.

dirk-thomas: Thanks for the fast update.

PR #86, opened 2016-08-30
change verify functions to skip deprecated stuff

dirk-thomas:
* By default don't try to verify the deprecated doc / release / source files.
* Disable the file providers for the build files if index is older then 3 since the index file won't have them and the call just results in a deprecation message.


tfoote:
+1

PR #95, opened 2017-02-21
split Debian package into scripts and modules

dirk-thomas:
Same as for [catkin-pkg](ros-infrastructure/catkin_pkg#157).

After this has been reviewed and merged I will release version 0.6.0.


dirk-thomas:
https://github.com/ros-infrastructure/rosdistro/pull/95/files#diff-419aa3e2c65a318d41b25af356c30f36R13 and following.

dirk-thomas:
The failing test is due to the GitHub API throttling the requests from Travis.

dirk-thomas:
For this to actually work `rospkg` needs to be split, too... :worried:

mikepurvis:
What is the objective here?

dirk-thomas:
Side by side installation of Python 2 and 3 modules, see the referenced catkin-pkg ticket for more information.

Issue #93, opened 2017-01-06
repository_specification.py does not recognize GitHub token URLs


ayrton04:
We're seeing a lot of messages like this in the rosdistro cache jobs:

```
DEBUG:rosdistro:Skipped "github_manifest_provider()": VCS url "https://<token>@github.com/blah/blah.git" does not match expected format.
```

I believe the offending regex is here:
https://github.com/ros-infrastructure/rosdistro/blob/2bb8505ec0bdfcc57d7b2a1033235eb3ce21f9f2/src/rosdistro/repository_specification.py#L40

dirk-thomas:
Your URL with the token is not one of the allowed URLs which are being tested in https://github.com/ros-infrastructure/rosdistro/blob/master/test/test_repository_specification.py Therefore I am not surprised that it doesn't pass the regex. Maybe @mikepurvis can comment on this since he added the stricter regex in #79.

mikepurvis:
That can be added to the regex without too much trouble. I didn't realize it was possible to use a personal access token like that with Github HTTPS URLs, but that seems a reasonable alternative to SSH urls. Perhaps better even, since it allows anyone with a copy of the rosdistro to get read access to all the repos without them needing to have anything explicitly set up.

The only hiccup is that adding a new match group for the token/username is going to change the tuple which comes back from `get_url_parts`, either that or we include the token as part of the "domain" portion of it, meaning that the manifest provider match checks need to change from equality to endswith or a regex: https://github.com/ros-infrastructure/rosdistro/search?q=get_url_parts

ayrton04:
Do you have a preference?

mikepurvis:
Changing the domain to include the token as required seems less disruptive to me. The main thing is only including it when it's actually a token and avoiding the `git@` that comes along with SSH URLs.

ayrton04:
Do we want to make it specific to the GH tokens, which are 40 hex characters? I think this should work:

```
(?:https?:\/\/|ssh:\/\/|git:\/\/|git@)(([a-fA-F0-9]{40}@)?[\w.-]+)[:/]([\w/-]*)(?:\.git)?$
```

mikepurvis:
Fine by me. Obviously update the test accordingly so we know what's actually going on, and update the instances of `get_url_parts`.

ayrton04:
:+1: Will do.

PR #97, opened 2017-04-04
Supporting GitHub auth tokens in manifest provider URLs

ayrton04:
Addresses #93.

### DIFF
--- a/test/test_repository_specification.py
+++ b/test/test_repository_specification.py
@@ -19,3 +19,9 @@ def test_repository_specification():
 
     r.url = 'git@example.com:a/b/c/d/e.git'
     assert r.get_url_parts() == ('example.com', 'a/b/c/d/e')
+
+    r.url = 'https://1234567890abcdeABCDE1234567890abcdeABCDE@example.com:a/b/c/d/e.git'
+    parts = r.get_url_parts()
+    assert(len(parts) == 2)
+    assert(parts[0].endswith('example.com'))
+    assert(parts[1] == 'a/b/c/d/e')
--- a/test/test_repository_specification.py
+++ b/test/test_repository_specification.py
@@ -22,6 +22,6 @@ def test_repository_specification():
 
     r.url = 'https://1234567890abcdeABCDE1234567890abcdeABCDE@example.com:a/b/c/d/e.git'
     parts = r.get_url_parts()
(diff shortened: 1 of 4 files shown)


test/test_repository_specification.py
```diff
@@ -19,3 +19,9 @@ def test_repository_specification():
 
     r.url = 'git@example.com:a/b/c/d/e.git'
     assert r.get_url_parts() == ('example.com', 'a/b/c/d/e')
+
+    r.url = 'https://1234567890abcdeABCDE1234567890abcdeABCDE@example.com:a/b/c/d/e.git'
+    parts = r.get_url_parts()
+    assert(len(parts) == 2)
+    assert(parts[0].endswith('example.com'))
+    assert(parts[1] == 'a/b/c/d/e')
```
dirk-thomas: `assert` is not a function but a statement. I should not have parenthesis.

dirk-thomas:
@mikepurvis Does this look good to you too?

test/test_repository_specification.py
```diff
@@ -19,3 +19,9 @@ def test_repository_specification():
 
     r.url = 'git@example.com:a/b/c/d/e.git'
     assert r.get_url_parts() == ('example.com', 'a/b/c/d/e')
+
+    r.url = 'https://1234567890abcdeABCDE1234567890abcdeABCDE@example.com:a/b/c/d/e.git'
+    parts = r.get_url_parts()
+    assert len(parts) == 2
+    assert parts[0].endswith('example.com')
+    assert parts[1] == 'a/b/c/d/e'
```
mikepurvis: Wouldn't we follow the pattern of the above asserts? It would be best to assert that it includes the token, no?

```
assert r.get_url_parts() == ('1234567890abcdeABCDE1234567890abcdeABCDE@example.com', 'a/b/c/d/e')
```

ayrton04: Yeah, that makes more sense. Will change.

mikepurvis:
LGTM. Test failures are unrelated— I think we've hit a limit on anonymous github API usage in a different test.

dirk-thomas:
Thank you!

dirk-thomas:
Released in version 0.6.2.

ayrton04:
Great, thanks!

PR #98, opened 2017-04-29
ignore_comments_in_yaml_file_comparison

mikaelarguedas:
this adds a parameter to ignore comments when comparing distribution / index files. this should allow https://github.com/ros/rosdistro/pull/14517 to be merged as  is

### DIFF
--- a/src/rosdistro/verify.py
+++ b/src/rosdistro/verify.py
@@ -140,16 +140,10 @@ def _check_file_identical(dist_file, yaml_url, file_type, ignore_comments=False)
     dist_file_lines[0:0] = _yaml_header_lines(file_type, dist_file_data['version'])
 
     if ignore_comments:
-        import re
-        yaml_lines_to_remove = []
         yaml_header = _yaml_header_lines(file_type, dist_file_data['version'])
-        for yaml_line in yaml_lines:
-            if yaml_line in yaml_header:
-                continue
-            elif re.match(r'(^#| +#)', yaml_line):
-                yaml_lines_to_remove.append(yaml_line)
-        for yaml_line_to_remove in yaml_lines_to_remove:
-            yaml_lines.remove(yaml_line_to_remove)
+        yaml_lines = [
+            line for line in yaml_lines
+            if line in yaml_header or not line.lstrip().startswith('#')]
 
     if yaml_lines != dist_file_lines:
         diff = difflib.unified_diff(
--- a/src/rosdistro/verify.py
+++ b/src/rosdistro/verify.py
@@ -106,7 +106,7 @@ def _reformat_files(index, dist_name, loader_function, yaml_url, file_type):
         if not url.startswith('file://'):
             print('Skipping non-file url: %s' % url)
             continue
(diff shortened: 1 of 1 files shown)


mikepurvis:
Strongly in support of this direction— however, note that without a wholesale switch to [ruamel.yaml](https://pypi.python.org/pypi/ruamel.yaml), other tools are still going to mangle/drop comments in `distribution.yaml` (eg Bloom PR creation; anything that uses the rosdistro python library).

mikaelarguedas:
> a wholesale switch to ruamel.yaml

Thanks for the pointer, I was looking for a parser implementation supporting comments!

> other tools are still going to mangle/drop comments in distribution.yaml (eg Bloom PR creation; anything that uses the rosdistro python library).

That is a very good point.

I don't know if @ros-infrastructure/ros_team would be willing to change the parser completely given that this will also impact any code interacting with distribution files. Maybe the simplest is to just remove the EOL distros instead of commenting them out as suggested [here](https://github.com/ros/rosdistro/pull/14517#issuecomment-294379292). Thoughts?

src/rosdistro/verify.py
```diff
@@ -128,17 +128,29 @@ def _check_files_identical(index, dist_name, loader_function, yaml_url, file_typ
...
+        import re
+        yaml_lines_to_remove = []
+        yaml_header = _yaml_header_lines(file_type, dist_file_data['version'])
+        for yaml_line in yaml_lines:
+            if yaml_line in yaml_header:
+                continue
+            elif re.match(r'(^#| +#)', yaml_line):
+                yaml_lines_to_remove.append(yaml_line)
+        for yaml_line_to_remove in yaml_lines_to_remove:
+            yaml_lines.remove(yaml_line_to_remove)
```
dirk-thomas: Could this logic be simplified with a list comprehension? E.g. like this (untested):

```
yaml_lines = [
    line for line in yaml_lines
    if line not in yaml_header and not line.lstrip().startswith('#')]
```

dirk-thomas:
I think the only tool in our pipeline writing the distribution files is `bloom` (which uses `rosdistro` for this internally?). Those would need to be updated to use a parser which keeps comments. I am not sure if it is worth the effort.

src/rosdistro/verify.py
mikaelarguedas: good point, now that I simplified the regex to not process the header but only the comment lstrip will totally do the job :+1:
fixed in https://github.com/ros-infrastructure/rosdistro/pull/98/commits/e1e84a0282445b8ee74099b9aa9c4b0b374daaf5

mikaelarguedas:
I don't think I'll put the effort in changing parser in the near future. Mostly because I think that relying on python core modules is preferred vs using third party packages. I modified https://github.com/ros/rosdistro/pull/14517 to remove the EOL distros rather than commenting them out. @ros-infrastructure/ros_team feel free to either close this or keep it open for visibility if we ever want to build on top.

dirk-thomas:
Since this would require the returned data structures to fundamentally change I don't think this will ever happen. So I will close it.

PR #100, opened 2017-05-26
Changed order of the url check.

### LINKED ISSUE #99: Legacy behavior during xml fetching
While working on the gentoo ebuild generation system, I found that there is always a request made
to the wrong url, perhaps for legacy reasons.

```
Failed to read package.xml file from url 'https://raw.github.com/ros-gbp/linux_peripheral_interfaces-release/release/libsensors_monitor/0.2.0/package.xml': HTTP Error 404: Not Found
Trying to read from url 'https://raw.github.com/ros-gbp/linux_peripheral_interfaces-release/release/kinetic/libsensors_monitor/0.2.0-0/package.xml' instead
```

The difference is `release/libsensors_monitor/0.2.0` vs. `release/kinetic/libsensors_monitor/0.2.0-0` (the later having a rosdistro affiliation, as well as a debian package increment.

allenh1:
Fixes #99

### DIFF
--- a/src/rosdistro/rosdistro.py
+++ b/src/rosdistro/rosdistro.py
@@ -199,24 +199,21 @@ class RosPackage:
         repo = self.repository
         if 'github.com' in repo.url:
             url = repo.url
-            upstream_version = repo.version.split('-')[0]
-            release_tag = 'release/{0}/{1}'.format(self.name, upstream_version)
-            url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
+            release_tag = 'release/{0}/{1}/{2}/package.xml'.format(rosdistro, self.name, repo.version)
             url = url.replace('git://', 'https://')
             url = url.replace('https://', 'https://raw.')
+            url = url.replace('.git', release_tag)
+
             try:
                 try:
                     package_xml = urlopen(url).read()
                 except Exception as e:
                     msg = "Failed to read package.xml file from url '{0}': {1}".format(url, e)
                     warning(msg)
-                    url = repo.url
-                    release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
-                    tail = '/{0}/package.xml'.format(release_tag)
-                    url = url.replace('.git', tail)
-                    url = url.replace('git://', 'https://')
-                    url = url.replace('https://', 'https://raw.')
-                    info("Trying to read from url '{0}' instead".format(url))
+                    upstream_version = repo.version.split('-')[0]
+                    legacy_release_tag = 'release/{0}/{1}/package.xml'.format(self.name, upstream_version)
+                    url = url.replace(release_tag, legacy_release_tag)
+                    info("Trying to read from legacy-style url '{0}' instead".format(url))
                     package_xml = urlopen(url).read()
             except Exception as e:
                 msg += '\nAND\n'
--- a/src/rosdistro/rosdistro.py
+++ b/src/rosdistro/rosdistro.py
@@ -203,7 +203,6 @@ class RosPackage:
             url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
             url = url.replace('git://', 'https://')
             url = url.replace('https://', 'https://raw.')
-
             try:
                 try:
                     package_xml = urlopen(url).read()
--- a/src/rosdistro/rosdistro.py
+++ b/src/rosdistro/rosdistro.py
@@ -199,10 +199,10 @@ class RosPackage:
         repo = self.repository
         if 'github.com' in repo.url:
             url = repo.url
-            release_tag = 'release/{0}/{1}/{2}/package.xml'.format(rosdistro, self.name, repo.version)
+            release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
+            url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
             url = url.replace('git://', 'https://')
             url = url.replace('https://', 'https://raw.')
-            url = url.replace('.git', release_tag)
 
             try:
                 try:
@@ -211,7 +211,7 @@ class RosPackage:
                     msg = "Failed to read package.xml file from url '{0}': {1}".format(url, e)
                     warning(msg)
                     upstream_version = repo.version.split('-')[0]
-                    legacy_release_tag = 'release/{0}/{1}/package.xml'.format(self.name, upstream_version)
+                    legacy_release_tag = 'release/{0}/{1}'.format(self.name, upstream_version)
                     url = url.replace(release_tag, legacy_release_tag)
                     info("Trying to read from legacy-style url '{0}' instead".format(url))
                     package_xml = urlopen(url).read()


src/rosdistro/rosdistro.py
```diff
@@ -204,20 +204,21 @@ def _fetch_package_xml(self, rosdistro):
...
                 try:
-                    package_xml = urlopen(url).read()
-                except Exception as e:
-                    msg = "Failed to read package.xml file from url '{0}': {1}".format(url, e)
-                    warning(msg)
                     url = repo.url
                     release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
                     tail = '/{0}/package.xml'.format(release_tag)
                     url = url.replace('.git', tail)
-                    url = url.replace('git://', 'https://')
```
mikaelarguedas: Has this been removed intentionally ? If yes, why did this become unnecessary ?

allenh1: It was not intentional. Good catch. Oddly, it works, but... I haven't tried it for indigo yet. One moment.

dirk-thomas:
My previous comment still applies: why shuffle all this code? Isn't it enough to swap these two lines: https://github.com/ros-infrastructure/rosdistro/blob/d5f6f0310bb6420945068aeb79d2f557a12a0951/src/rosdistro/rosdistro.py#L203 and https://github.com/ros-infrastructure/rosdistro/blob/d5f6f0310bb6420945068aeb79d2f557a12a0951/src/rosdistro/rosdistro.py#L214

src/rosdistro/rosdistro.py
```diff
@@ -198,25 +198,26 @@ def __init__(self, name, repository):
...
+                    package_xml = urlopen(url).read()
+                except Exception as e:                    
+                    msg = "Failed to read package.xml file from url '{0}': {1}".format(url, e)
+                    warning(msg)
+                    url = repo.url
+                    upstream_version = repo.version.split('-')[0]
+                    release_tag = 'release/{0}/{1}'.format(self.name, upstream_version)
+                    url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
+                    url = url.replace('git://', 'https://')
+                    url = url.replace('https://', 'https://raw.')
```
mikepurvis: This seems like a duplication of the more robust logic which is here:

https://github.com/ros-infrastructure/rosdistro/blob/master/src/rosdistro/manifest_provider/github.py#L65

Can we consolidate?

allenh1: That would be a much better solution, yes. Let me do that.

allenh1:
Upon further inspection, it appears that, in the function you linked, they call `get_release_tag` which calls
`fetch package url`.

https://github.com/ros-infrastructure/rosdistro/blob/master/src/rosdistro/rosdistro.py#L238

There seems to be some redundancy here in fetching the `package.xml` file. In particular, it happens twice. So I think the best consolidation here would be to modify the `github_manifest_provider` to not fetch the `package.xml` file, as it is already available.

mikepurvis:
In that case you'd want to use the CachedManifestProvider, as in:

https://github.com/ros-infrastructure/rosdistro/blob/d5f6f0310bb6420945068aeb79d2f557a12a0951/src/rosdistro/__init__.py#L141-L144

However, looks like it might be some crazy legacy-supporting side logic, so let's wait for a verdict from @dirk-thomas about the right path forward before you try to change too much.

allenh1:
Will do. I was mistaken by the structure a little -- I will play around with it some, then bother dirk in a few.

src/rosdistro/rosdistro.py
dirk-thomas: Please keep the actual change swapping the order or urls fetched separate from code refactoring (like deduplicating logic). At least in separate commits - maybe even in separate PRs.

allenh1:
@dirk-thomas Should be the correct format now.

@mikepurvis After digging through things for a good while, I've determined that the amount of refactoring needed to avoid duplication of this logic reaches beyond the bounds of this pull request.

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,24 @@ def _fetch_package_xml(self, rosdistro):
         repo = self.repository
         if 'github.com' in repo.url:
             url = repo.url
-            upstream_version = repo.version.split('-')[0]
-            release_tag = 'release/{0}/{1}'.format(self.name, upstream_version)
-            url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
+            # release_tag = get_release_tag(self.name)
```
dirk-thomas: This line should be removed.

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,24 @@ def _fetch_package_xml(self, rosdistro):
...
+            tail = '/{0}/package.xml'
+            url = url.replace('.git', tail)
             url = url.replace('git://', 'https://')
             url = url.replace('https://', 'https://raw.')
+
             try:
                 try:
+                    release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
+                    url = url.format(release_tag)
```
dirk-thomas: Before `format` was only called on strings defined in the source code. Now `format` is being called on a user provided string which might contain characters like `{}` and therefore result in problems. Therefore please only use `format` on strings defined in the code directly.

Same below in line 218:

> url = url.format(release_tag)

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,24 @@ def _fetch_package_xml(self, rosdistro):
...
                     package_xml = urlopen(url).read()
-                except Exception as e:
+                except Exception as e:
```
dirk-thomas: Please remove the trailing whitespace.

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,24 @@ def _fetch_package_xml(self, rosdistro):
...
-            url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
+            # release_tag = get_release_tag(self.name)
+            tail = '/{0}/package.xml'
+            url = url.replace('.git', tail)
             url = url.replace('git://', 'https://')
             url = url.replace('https://', 'https://raw.')
+
             try:
                 try:
+                    release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
```
dirk-thomas: This line as well as the next one shouldn't be in the `try` / `except` block.

allenh1:
@dirk-thomas I restructured the logic a little to fix it. Should be good now.

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,21 @@ def _fetch_package_xml(self, rosdistro):
         repo = self.repository
         if 'github.com' in repo.url:
             url = repo.url
-            upstream_version = repo.version.split('-')[0]
-            release_tag = 'release/{0}/{1}'.format(self.name, upstream_version)
-            url = url.replace('.git', '/{0}/package.xml'.format(release_tag))
+            tail = 'release/{0}/{1}/{2}/package.xml'.format(rosdistro, self.name, repo.version)
```
dirk-thomas: I would suggest to keep the variable name as `release_tag` since the string is the release tag.

src/rosdistro/rosdistro.py
```diff
@@ -199,24 +199,21 @@ def _fetch_package_xml(self, rosdistro):
...
                     warning(msg)
-                    url = repo.url
-                    release_tag = 'release/{0}/{1}/{2}'.format(rosdistro, self.name, repo.version)
-                    tail = '/{0}/package.xml'.format(release_tag)
-                    url = url.replace('.git', tail)
-                    url = url.replace('git://', 'https://')
-                    url = url.replace('https://', 'https://raw.')
-                    info("Trying to read from url '{0}' instead".format(url))
+                    upstream_version = repo.version.split('-')[0]
+                    legacy_tail = 'release/{0}/{1}/package.xml'.format(self.name, upstream_version)
```
dirk-thomas: I would suggest to rename the variable to `legacy_release_tag`.

allenh1: @dirk-thomas you got it

dirk-thomas:
The previous state seem to lack the leading slash in the replacement: see https://github.com/ros-infrastructure/rosdistro/pull/100/commits/1c6eb29c650925b74411f67f62a1377dd2100a34#diff-65e54a8cd0a3220cd2b0b2091bd58708R202

Anyway I added two commits to this PR to minimize the changes and keep the semantic of the variables (`release_tag` and `legacy_release_tag` are actualy containing valid tag names of the repo). Please double check that the overall patch still works as expected.

allenh1:
@dirk-thomas All is working as expected!

dirk-thomas:
Thanks!

PR #103, opened 2017-06-16
Enable github private repos.

jgoppert:
Needs: https://github.com/ros-infrastructure/bloom/pull/429

> I'm making a private build farm, and I've seen a lot of people doing local clones and file:/// references for their ROS_DISTRO_URL and this was just so horribly clunky I fixed it.
> 
> This could probably be better intertwined with the .config/bloom settings. I had to set both the token in .config/bloom and GITHUB_PASSWORD to my github "Persontal Access Token" to get it to work.
>
> This also supports private automatic PR's from bloom!

### DIFF
--- a/src/rosdistro/loader.py
+++ b/src/rosdistro/loader.py
@@ -33,19 +33,46 @@
 
 import socket
 import time
+import os
+import sys
+import base64
+
+from . import logger
+
 try:
-    from urllib.request import urlopen
+    from urllib.request import urlopen, Request
     from urllib.error import HTTPError
     from urllib.error import URLError
 except ImportError:
-    from urllib2 import urlopen
+    from urllib2 import urlopen, Request
     from urllib2 import HTTPError
     from urllib2 import URLError
 
 
+GITHUB_USER = os.getenv('GITHUB_USER', None)
+GITHUB_PASSWORD = os.getenv('GITHUB_PASSWORD', None)
+
+
+def auth_header(username=None, password=None, token=None):
+    if username and password:
+        if sys.version_info > (3, 0):
+            userandpass = base64.b64encode(bytes('%s:%s' % (username, password), 'utf-8'))
+        else:
+            userandpass = base64.b64encode('%s:%s' % (username, password))
+        userandpass = userandpass.decode('ascii')
+        return 'Basic %s' % userandpass
+    elif token:
+        return 'token %s' % token
+
+
 def load_url(url, retry=2, retry_period=1, timeout=10, skip_decode=False):
+    req = Request(url)
+    if GITHUB_USER and GITHUB_PASSWORD:
+        logger.debug('- using http basic auth from supplied environment variables.')
+        authheader = auth_header(username=GITHUB_USER, password=GITHUB_PASSWORD)
+        req.add_header('Authorization', authheader)
     try:
-        fh = urlopen(url, timeout=timeout)
+        fh = urlopen(req, timeout=timeout)
     except HTTPError as e:
         if e.code in [500, 502, 503] and retry:
             time.sleep(retry_period)
--- a/src/rosdistro/manifest_provider/bitbucket.py
+++ b/src/rosdistro/manifest_provider/bitbucket.py
@@ -73,7 +73,7 @@ def bitbucket_manifest_provider(_dist_name, repo, pkg_name):
         req = Request(url)
         if BITBUCKET_USER and BITBUCKET_PASSWORD:
             logger.debug('- using http basic auth from supplied environment variables.')
-            authheader = 'Basic %s' % base64.b64encode('%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
+            authheader = 'Basic %s' % base64.b64encode(b'%s:%s' % (BITBUCKET_USER, BITBUCKET_PASSWORD))
             req.add_header('Authorization', authheader)
         package_xml = urlopen(req).read().decode('utf-8')
         return package_xml
--- a/src/rosdistro/manifest_provider/github.py
+++ b/src/rosdistro/manifest_provider/github.py
@@ -82,7 +82,7 @@ def github_source_manifest_provider(repo):
     req = Request(tree_url)
     if GITHUB_USER and GITHUB_PASSWORD:
         logger.debug('- using http basic auth from supplied environment variables.')
-        authheader = 'Basic %s' % base64.b64encode('%s:%s' % (GITHUB_USER, GITHUB_PASSWORD))
+        authheader = 'Basic %s' % base64.b64encode(b'%s:%s' % (GITHUB_USER, GITHUB_PASSWORD))
         req.add_header('Authorization', authheader)
     try:
         tree_json = json.loads(urlopen(req).read().decode('utf-8'))


jgoppert:
Not sure what this build error is related to:

```bash
RuntimeError: Unable to fetch JSON tree from https://api.github.com/repos/ros/ros/git/trees/kinetic-devel?recursive=1: HTTP Error 403: Forbidden
```

mikepurvis:
At Clearpath, we use SSH URLs, and that works totally fine with Github, Bitbucket, and GitLab.

Additionally, you can put auth tokens in your URLs, as of #93.

jgoppert:
@mikepurvis I tried using ssh urls and it sort of works, but the issue is that any reference to urls like: https://raw.githubusercontent.com/ros/rosdistro/master/index.yaml in private repos, don't work, even if you put the token.

jgoppert:
@mikepurvis @dirk-thomas any chance of getting this in? It is currently blocking my approach because the docker image doesn't have my patches to the ros python package since it is installing them via pip.  I guess I could find the docker script and modify it to checkout my branch, but seems like a lot of work.

If there is an alternative to this with ssh that doesn't require downloading your rosdistro and ros_buildfarm_config  locally with a file:/// url, let me know. But I have tried just putting the access tokens in the urls and that doesn't work with https://TOKEN@raw.githubusercontent.com/ros/rosdistro/master/index.yaml  as stated above.
Also, using ssh doesn't seem to support the automatic PR mechanism due to a similar issue with the github api.

jgoppert:
Currently stuck here, the docker image uses sudo apt-get install python3-rosdistro-modules, which doesn't have my patch yet.

```bash
# END SECTION
# BEGIN SECTION: Run Dockerfile - generate sourcedeb
# BEGIN SUBSECTION: get sources
# END SUBSECTION
  File "/tmp/ros_buildfarm/scripts/release/get_sources.py", line 50, in <module>
    sys.exit(main())
  File "/tmp/ros_buildfarm/scripts/release/get_sources.py", line 46, in main
    args.os_name, args.os_code_name, args.source_dir, args.debian_repository_urls)
  File "/tmp/ros_buildfarm/ros_buildfarm/sourcedeb_job.py", line 28, in get_sources
    index = get_index(rosdistro_index_url)
  File "/usr/lib/python3/dist-packages/rosdistro/__init__.py", line 101, in get_index
    yaml_str = load_url(url)
  File "/usr/lib/python3/dist-packages/rosdistro/loader.py", line 48, in load_url
    fh = urlopen(url, timeout=timeout)
  File "/usr/lib/python3.5/urllib/request.py", line 163, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.5/urllib/request.py", line 472, in open
    response = meth(req, response)
  File "/usr/lib/python3.5/urllib/request.py", line 582, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib/python3.5/urllib/request.py", line 510, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.5/urllib/request.py", line 444, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.5/urllib/request.py", line 590, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found (https://raw.githubusercontent.com/MY_REPO/rosdistro/master/index.yaml)
```

jgoppert:
@mikepurvis I've got a temporary fix working. I can host my rosdistro, ros_buildfarm_config, and apt repository locally via the docker address for localhost: http://172.17.0.1/  Then I can use my patched ros distro tools to create secure PR's on github with bloom, so no rush, but still would be good to get this in for others to use.

src/rosdistro/loader.py
```diff
@@ -33,19 +33,46 @@
...
+    from urllib2 import urlopen, Request
     from urllib2 import HTTPError
     from urllib2 import URLError
 
 
+GITHUB_USER = os.getenv('GITHUB_USER', None)
+GITHUB_PASSWORD = os.getenv('GITHUB_PASSWORD', None)
+
+
+def auth_header(username=None, password=None, token=None):
```
mikepurvis: I don't understand what's going on in this function, and in particular why it only needs to happen here and not for the manifest providers as well. Should this be a generic thing that's used everywhere?

src/rosdistro/loader.py
```diff
@@ -33,19 +33,46 @@
...
+            userandpass = base64.b64encode('%s:%s' % (username, password))
+        userandpass = userandpass.decode('ascii')
+        return 'Basic %s' % userandpass
+    elif token:
+        return 'token %s' % token
+
+
 def load_url(url, retry=2, retry_period=1, timeout=10, skip_decode=False):
+    req = Request(url)
+    if GITHUB_USER and GITHUB_PASSWORD:
```
mikepurvis: I don't care for passing my github credentials to an arbitrary URL— shouldn't we check here if the request URL is in fact Github before doing this?

Alternatively, maybe some of this logic should be hoisted into the `auth_header` function; pass it the URL and have it do the business of checking the host and then pulling the credentials or token from the environment as appropriate.

src/rosdistro/loader.py
jgoppert: It is just creating either a basic or token auth header. I had to have the same function in bloom and ros_biuldfarm as well. So if there is one module that all of these import, we can put it there.

src/rosdistro/loader.py
jgoppert: GITHUB_USER is a bit mis-leading, this also works for bitbucket as far as I can tell from the existing usage of it, but it seems that naming convention was left over. https://developer.atlassian.com/bitbucket/server/docs/latest/how-tos/example-basic-authentication.html

The issue would be if someone was using gitlab. If we wanteed to do something like bloom does and keep the settings in a config file, we could have user enter auth for each url in a dict.

```JSON
{
"https://www.github.com/my_company/my_repo": ["user", "password"]
}
```

Or something like that.

mikepurvis: Right, but for my use case, I have repos at Github, Bitbucket, and Gitlab, and I have different credentials for all three in the environment that fetches manifests to build the cache.

I would prefer if this logic followed the same pattern and used the correct set of credentials based on where he rosdistro is located.

jgoppert: Write, I'm suggesting we do something like:

cat .config/rosdistro
{
"https://www.github.com/my_company/my_repo": ["user", "password"],
"https://www.github.com/my_company/my_repo": [user="user", password="password"],
"https://www.mygitlab.com/my_team/my_repo": [user="user", password="password"],
"https://www.github.com/my_company/my_repo2": [token="token"],
}

mikepurvis: I'm not seeing a motivation to depart from the convention already established in this repo.

jgoppert: OK, are automated  pull requests working for you? I noticed most of it was broken for private repo and did all of this to get it working. Am I missing something?  It would be good to add secure pr's to the unit tests to get coverage over this code.

jgoppert: And what is the current method of handling raw files from github lIke for distrouble yaml. I don't think there is any, correct?

mikepurvis: The existing convention works for building the rosdistro cache from private repos on Github and Bitbucket, and is implemented in the manifest providers:

https://github.com/ros-infrastructure/rosdistro/tree/master/src/rosdistro/manifest_provider

The envvars in question are `GITHUB_USER`, `GITHUB_PASSWORD`, `BITBUCKET_USER`, and `BITBUCKET_PASSWORD`. My request is that we use these same envvars for this new logic, but use the _correct_ ones depending on where the rosdistro is hosted.

jgoppert: Ah, OK, sorry for misunderstanding. I hadn't seen BITBUCKET_USER and BITBUCKET_PASSWORD yet.

mikepurvis: My apologies also; I should have been more explicit upfront.

dirk-thomas:
@jgoppert What is the status of this PR? Is it obsolete and can be closed since the environment variables are sufficient?

dirk-thomas:
I will go ahead and close this PR due to the long time of inactivity. Please feel free to comment and if necessary the ticket can be reopened.

lsolanka-flya:
> At Clearpath, we use SSH URLs, and that works totally fine with Github, Bitbucket, and GitLab.
> 
> Additionally, you can put auth tokens in your URLs, as of #93.

Hi, could you maybe provide an example of such an URL to download the index file from Bitbucket? I am trying to set up a private ROS buildfarm but it looks like I'm missing some (trivial?) way how to access an `index.yaml` file from a private rosdistro clone on bitbucket.

mikepurvis:
@lsolanka-flya There may well be some way to do this, but if so I'm not sure what it is— we've always had our private rosdistro repo world-readable behind our firewall.

lsolanka-flya:
@mikepurvis sorry to hijack this thread, but if you were to set up a buildfarm with rosdistro as a private repo, would you see it as a realistic solution? I see a lot of code in ros-infrastructure/ros_buildfarm that depends on various configuration repositories (such as rosdistro, ros_buildfarm_configuration, etc.?) to be publicly available through https so I'm wondering whether it's even worth digging into it. Thanks for an answer.

gavanderhoorn:
> I see a lot of code in ros-infrastructure/ros_buildfarm that depends on various configuration repositories (such as rosdistro, ros_buildfarm_configuration, etc.?) to be publicly available through https

a nuance here: while I agree it would be good to have support for private repositories, it would seem the actual requirement here is "repositories that can only be accessed after authenticating".

Github only has that for its *private* repositories, but that is just one implementation.

There is no requirement in the packages that you list for repositories to be "public", but it is easier if those repositories are accessible without requiring authentication (note the difference).

I believe that is what @mikepurvis implies: repositories can be made available without requiring authentication by placing them behind a firewall (or some other form of network isolation). This would allow your company/organisation/buildfarm to access them, but not anyone (or anything) else.

lsolanka-flya:
> > I see a lot of code in ros-infrastructure/ros_buildfarm that depends on various configuration repositories (such as rosdistro, ros_buildfarm_configuration, etc.?) to be publicly available through https
> 

> I believe that is what @mikepurvis implies: repositories can be made available without requiring authentication by placing them behind a firewall (or some other form of network isolation). This would allow your company/organisation/buildfarm to access them, but not anyone (or anything) else.

@gavanderhoon thanks for the reply. However that to me implies that also my buildfarm is generally behind that firewall, which is not the case I'm trying to set up. But perhaps I'm missing something here...

gavanderhoorn:
I don't believe that has to be the case.

Firewalls can be configured to allow (and deny) all sorts of traffic through or block it. You could place your configuration repositories behind a firewall, make sure your buildfarm only has access to it, and still allow "the world" to access your buildfarm `master` host.

You'd have to make sure to not allow unauthenticated users access to your job configuration, but that would keep everything relatively secure (or at least: as secure as a setup *with* authentication for repositories would seem to be).

Perhaps you could detail *which* parts of your buildfarm you'd want to expose. If it's just the `repo` host, then that would seem to be trivial. It's just a web server that serves `.deb`s.

lsolanka-flya:
Yeah, no problem, let me just set the stage a little (and thanks for you time!). Also I have to say I'm not an expert on the networking infrastructure so I might be quite wrong on some things.

I am trying to set up a completely private buildfarm in my company (including the apt repository), entirely on the cloud as I don't have in-house resources to maintain the farm (nor to set up firewall access and external access such as VPN if needed). The jenkins part would be hosted by AWS and the apt repository server I guess also on AWS. All sources are in private repos (Github/Bitbucket), including forks of https://github.com/ros/rosdistro and https://github.com/ros-infrastructure/ros_buildfarm_config.

Here I hit the roadblock I am describing, since to me world-readable in this configuration means also public.

gavanderhoorn:
If you're hosting your repositories on networks that are not under your control (and using Github and Bitbucket would fall in that category) then it would seem you can't use the setup @mikepurvis describes.

I'm guessing clearpath has an EE deployment of either gitlab or github (or some other platform) which is on-premise or hosted in a cloud environment they can control.

mikepurvis:
If you don't like having to mirror the rosdistro locally, another option might be to throw a proxy in front of it that takes care of the auth for you. An example of this setup is here: https://stackoverflow.com/a/571260

Admittedly, this is still a hack, and I can definitely understand wanting a more native/transparent solution for this.

Issue #105, opened 2017-11-14
Reverse depends on not working for Source Dependency Walker


tspicer01:
The get_depends_on and get_recursive_depends_on are not working for a SourceDependencyWalker.
The functions are trying to pull the repo from the release_repository which in our case doesn't exist as we have only the source.

Proposal
Have a new function _get_repo that can be used internally and overridden in the SourceDependencyWalker to return the repo depending on the object that is being used.  This will allow for the SourceDependencyWalker to use the parent functions.

dirk-thomas:
Addressed by #106.

PR #106, opened 2017-11-14
Fix get_depends_on to work for Source Walker

tspicer01:
Fix get_depends_on to work with a SourceDependencyWalker.

Added a _get_repo function to both DependencyWalker and SourceDependencyWalker so that the function can just just call the appropriate objects function to get the repo.

This will fix issue #105

### DIFF
--- a/src/rosdistro/dependency_walker.py
+++ b/src/rosdistro/dependency_walker.py
@@ -93,7 +93,7 @@ class DependencyWalker(object):
         for name in self._get_package_names():
             if name in ignore_pkgs:
                 continue
-            repo = self._get_repo(name)
+            repo = self._get_package_repo(name)
             if repo is None or repo.version is None:
                 continue
             deps = self._get_dependencies(name, depend_type)
@@ -110,6 +110,7 @@ class DependencyWalker(object):
             next_pkg_to_check = pkgs_to_check.pop()
             for depend_type in depend_types:
                 deps = self.get_depends_on(next_pkg_to_check, depend_type, ignore_pkgs=ignore_pkgs)
+x
                 new_deps = deps - depends_on
                 pkgs_to_check |= new_deps
                 depends_on |= new_deps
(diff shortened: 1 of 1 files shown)


mikepurvis:
Looks like a nice clean fix. I'm surprised we didn't run into this oversight sooner, whoops. :)

👍

src/rosdistro/dependency_walker.py
```diff
@@ -93,8 +93,7 @@ def get_depends_on(self, pkg_name, depend_type, ignore_pkgs=None):
         for name in self._get_package_names():
             if name in ignore_pkgs:
                 continue
-            pkg = self._get_package(name)
-            repo = self._distribution_instance.repositories[pkg.repository_name].release_repository
```
mikepurvis: I think the `pkg.repository_name` logic should either be maintained, or the new function should be `_get_repo_for_package`, since right now it's a bit ambiguous whether it's "get the repo named X" or "get the repo for package named X".

tspicer01: Valid point on the function name, I've changed it.

The reason that I changed how the release repo was determining the repo was b/c on my system it was continually failing with repository_name not a valid.  I dumped the pkg object and don't see repository_name being set.
I also checked in catkin_pkg parse_package_string and don't see it defined.  

If this is an issue with my version compatibility I can change it back but I'm unsure of how it's currently working for the normal DependencyWalker use case.

mikepurvis: It's set here: https://github.com/ros-infrastructure/rosdistro/blob/fc9b2441c5d79b918164c62a0f470f0934a6b172/src/rosdistro/package.py#L39

The `Package` object in question is from `catkin_pkg`; it's a structure internal to `rosdistro`. Anyway, changing the package name is totally satisfactory to me.

dirk-thomas:
Thank you for the fix.

dirk-thomas:
Available in version 0.6.3.

PR #107, opened 2017-12-28
Add retry on git remote

tspicer01:
We have seen some intermittent issues with ls-remote calls which cause the version not to be frozen.
Introducing a retry on the git ls-remote call if it fails the first time to handle these intermittent connection problems.

### DIFF
--- a/src/rosdistro/freeze_source.py
+++ b/src/rosdistro/freeze_source.py
@@ -83,33 +83,25 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
 
 
 # Get the repo commit information
-def _get_repo_info(source_repo, retry_count=3, sleep_time=1):
-    count = 1
-    # increase retry count to handle initial call
-    retry_count += 1
-    # allow for a retry if the ls-remote fails to query the endpoint repo
-    while True:
-        try:
-            cmd = ['git', 'ls-remote', source_repo.url]
-            ls_remote_lines = subprocess.check_output(cmd).splitlines()
-            return ls_remote_lines
-        except subprocess.CalledProcessError as err:
-            print("Non-zero return code for: %s, attempt %s of %s, sleeping for %s seconds and retrying command" %
-                 (' '.join(cmd), count, retry_count, sleep_time), file=sys.stderr)
-            count += 1
-            # reraise the error if we've attempted retry_count times
-            if count > retry_count:
-                err.message += " - Attempted %s time(s)" % retry_count
-                raise
-            # brief delay incase its an intermittent issue with infrastructure
-            time.sleep(sleep_time)
+def _get_repo_info(url, retry=2, retry_period=1):
+    cmd = ['git', 'ls-remote', url]
+    try:
+        return subprocess.check_output(cmd).splitlines()
+    except subprocess.CalledProcessError as err:
+        if not retry:
+            raise
+        print('  Non-zero return code for: %s, retrying in %f seconds' %
+              (' '.join(cmd), retry_period), file=sys.stderr)
+        # brief delay incase its an intermittent issue with infrastructure
+        time.sleep(retry_period)
+        return _get_repo_info(source_repo, retry=retry - 1, retry_period=retry_period * 2)
 
 
 def _worker(work_queue):
     while True:
         try:
             source_repo, freeze_version, freeze_to_tag = work_queue.get(block=False)
-            ls_remote_lines = _get_repo_info(source_repo)
+            ls_remote_lines = _get_repo_info(source_repo.url)
             for line in ls_remote_lines:
                 hash, ref = line.split('\t', 1)
                 if freeze_to_tag and ref == 'refs/tags/%s' % freeze_version:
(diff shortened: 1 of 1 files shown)


src/rosdistro/freeze_source.py
```diff
@@ -100,7 +118,7 @@ def _worker(work_queue):
             work_queue.task_done()
 
         except subprocess.CalledProcessError:
-            print("Non-zero return code for: %s" % ' '.join(cmd), file=sys.stderr)
+            print("No information could be retrieved for repo %s" % source_repo.url)
```
mikepurvis: Right now we just spew this error to the console but otherwise behave as if successful. A further improvement at some point could track these failures and perhaps abort the overall operation, returning an error code.

src/rosdistro/freeze_source.py
```diff
@@ -82,12 +82,30 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
...
         try:
-            source_repo, freeze_version, freeze_to_tag = work_queue.get(block=False)
             cmd = ['git', 'ls-remote', source_repo.url]
             ls_remote_lines = subprocess.check_output(cmd).splitlines()
+            return ls_remote_lines
+        except:
+            print("Non-zero return code for: %s retrying command" % ' '.join(cmd), file=sys.stderr)
+            count += 1
+            # reraise the error if we've attempted 3 times
+            if count > 3:
```
tfoote: It would be better if the hard coded retry count and sleep duration were default parameters to the function. They'll be much easier to find there when reading the code.

src/rosdistro/freeze_source.py
```diff
@@ -82,12 +82,30 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
...
+                raise
```
tfoote: It would be good to update the exception to indicate how many retries were attempted.

src/rosdistro/freeze_source.py
```diff
@@ -82,12 +82,30 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
...
+    count = 0
+    # allow for a retry if the ls-remote fails to query the endpoint repo
     while True:
         try:
-            source_repo, freeze_version, freeze_to_tag = work_queue.get(block=False)
             cmd = ['git', 'ls-remote', source_repo.url]
             ls_remote_lines = subprocess.check_output(cmd).splitlines()
+            return ls_remote_lines
+        except:
+            print("Non-zero return code for: %s retrying command" % ' '.join(cmd), file=sys.stderr)
```
tfoote: It would be good to include the sleep period in this message.

src/rosdistro/freeze_source.py
```diff
@@ -99,8 +119,9 @@ def _worker(work_queue):
 
             work_queue.task_done()
 
-        except subprocess.CalledProcessError:
-            print("Non-zero return code for: %s" % ' '.join(cmd), file=sys.stderr)
+        except subprocess.CalledProcessError as err:
+            print("No information could be retrieved for repo %s with error: %s %s" %
+                 (source_repo.url, err, err.message))
```
dirk-thomas: This message should go to `stderr`.

src/rosdistro/freeze_source.py
```diff
@@ -82,12 +82,32 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
         print("")
 
 
-def _worker(work_queue):
+# Get the repo commit information
+def _get_repo_info(source_repo, retry_count=3, sleep_time=1):
```
dirk-thomas: The current argument `retry_count` is misleading imo. I would expect that after the first attempt it will be *retried* N times but that doesn't seem to be the case. It is only being retried N-1 times for an overall number of N attempts.

Therefore I would suggest to change the logic so that it matches the meaning of "retry". Passing `0` should try once, passing `1` should try up to two times, etc.

tspicer01: Good catch. I had changed it so that the error message made more sense vs having it say Attempt 0 of x.  I've increased the retry count to account for the initial attempt.

dirk-thomas:
Thank you for the patch. I simplified the logic quite a bit (sacrificing the retry count in the error message) but adding an incrementally increasing sleep duration.

tspicer01:
@dirk-thomas   I appreciate the simplification work that you did.  I've been testing the new version and when it fails and does a retry it's now throwing an exception as source_repo is not defined.

```
Exception in thread Thread-16:
File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
File "/usr/lib/python2.7/threading.py", line 763, in run
self.__target(*self.__args, **self.__kwargs)
File "/var/lib/jenkins-slave/agent/agent/workspace/2.9-devel-xenial-bundle-starter/env/local/lib/python2.7/site-packages/rosdistro/freeze_source.py", line 104, in _worker
ls_remote_lines = _get_repo_info(source_repo.url)
File "/var/lib/jenkins-slave/agent/agent/workspace/2.9-devel-xenial-bundle-starter/env/local/lib/python2.7/site-packages/rosdistro/freeze_source.py", line 97, in _get_repo_info
return _get_repo_info(source_repo, retry=retry - 1, retry_period=retry_period * 2)
NameError: global name 'source_repo' is not defined
```

dirk-thomas:
There was an unrelated fix related to bootstrapping Melodic (#113) so I went ahead and merged the PR and made a release. Please let us know if version 0.6.6 works for your use case.

tspicer01:
Thanks @dirk-thomas  I tested version 0.6.6 for the specific scenario I was hitting and it worked.  The retry on the bad ls-remote allowed the repo to be frozen properly

dirk-thomas:
@tspicer01 Thank you for reporting back.

PR #108, opened 2018-01-05
Add bionic.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


tfoote:
We're getting 403 permission denied errors in the CI `RuntimeError: Unable to fetch JSON tree from https://api.github.com/repos/ros/genmsg/git/trees/0.5.7?recursive=1: HTTP Error 403: Forbidden`

`

clalancette:
Yeah, it's weird because it seems to work locally for me.  I'm looking into it now.

mikepurvis:
It's because that API has a rate-limiter on Github's end. If you don't supply a token, the limit is really low. Potentially someone could generate an actual API token and encrypt it for travis's use?

mikepurvis:
Looks like it's actually set up right now just to use a username and password, actually:

https://github.com/ros-infrastructure/rosdistro/blob/efcaaa46ffae16b450849c4f3a147bc60aac8cb1/src/rosdistro/manifest_provider/github.py#L49-L72

Either way, this is the mechanism for giving travis a secret to use: https://docs.travis-ci.com/user/encryption-keys/

clalancette:
> Looks like it's actually set up right now just to use a username and password, actually:

Even worse, the user/pass combination is only implemented for fetching source manifests, not package manifests.

Thanks for the pointer to the travis documentation; I'll take a look at that.

mikepurvis:
The API limiter issue only applies to the source manifest fetcher, since the other one is just accessing non-limited public URLs. I believe users with private repos either use SSH URLs or URLs with integrated tokens, so there's been no need to pass the github credentials on the other side.

clalancette:
After thinking about this more, I think I'm just going to close this out and re-open to see if we can get Travis to pass for now.  While I think the improvements that @mikepurvis suggested are valuable, they should probably be done on another PR (and we can hopefully get Melodic moving).  I'll open an issue about those improvements.

clalancette:
Sigh.  That didn't work.

I've looked into using the encrypted environment variable, which I think would work.  However, there is this statement in the Travis [documentation](https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions):

```
Pull requests sent from forked repositories do not have access to encrypted variables or data.
```

They then go on to say that you must skip tests when a PR comes from forked PRs anyway.  Because we have to do it anyway, I'm going to look at mocking out the API response for the test.

mikepurvis:
Yeah, just mock it out. That's what I should have done from the beginning— sorry for the hassle.

clalancette:
The latest commit adds mocking by monkey-patching `urlopen`, but it is really pretty terrible.  At the moment, I don't have a lot of better ideas short of switching us to another unit-test framework that has built-in monkey-patching abilities.  I'm definitely open to other ideas.

test/test_manifest_providers.py
```diff
@@ -46,9 +49,58 @@ def test_git_source():
...
+    # In both cases, store the file-like object as a module property so that we
+    # can properly close it during 'unmock_urlopen'.
+    if re.search('.*package.xml$', haystack) is not None:
+        sys.modules[__name__].mockfp = open('test/github-genmsg-package.xml', 'rb')
+    else:
+        sys.modules[__name__].mockfp = open('test/github-tree-data.json', 'rb')
+    return sys.modules[__name__].mockfp
+
+
+def mock_urlopen():
```
mikepurvis: Fine with me, but are we opposed to just depending on `python-mock` for this? v1.3.2 is present in Xenial.

mikepurvis:
Inside the target file, `urlopen(url).read().decode('utf-8')` is called three times. What about just putting a function in there which encapsulates that, and then you can just patch the wrapper function in the test? Might be lighter-weight than trying to patch `urlopen` itself?

```
def get_url_contents(url):
    urlopen(url).read().decode('utf-8')
```

Then in the test do a context manager or decorator or something:

```
import rosdistro.manifest_provider.github

def mock_url_contents(content):
    old_f = rosdistro.manifest_provider.github.get_url_contents
    rosdistro.manifest_provider.github.get_url_contents = lambda x: return content
    yield
    rosdistro.manifest_provider.github.get_url_contents = old_f

def test_foo():
    with mock_url_contents(canned_content):
        # test stuff
```

test/test_manifest_providers.py
clalancette: I'm certainly not opposed to using `python-mock`, though I have yet to be able to make it work :).  I'll try a bit more, just to see if it is nicer.

clalancette:
> Inside the target file, urlopen(url).read().decode('utf-8') is called three times. What about just putting a function in there which encapsulates that, and then you can just patch the wrapper function in the test? Might be lighter-weight than trying to patch urlopen itself?

I ended up doing a combination of this one, and using `python-mock` to monkey patch things.  It is *much* nicer now; thanks for the advice.  I think I'm pretty happy with how this is looks now; let me know what you think.

test/test_manifest_providers.py
```diff
@@ -46,9 +49,46 @@ def test_git_source():
...
+
+    # Now see if the URL ends with 'package.xml'.  If it does, open up the file
+    # that has the canned package.xml data we have for genmsg.  Otherwise,
+    # return the canned JSON tree data that github would have returned.
+    # In both cases, store the file-like object as a module property so that we
+    # can properly close it during 'unmock_urlopen'.
+    if re.search('.*package.xml$', haystack) is not None:
+        fname = 'test/github-genmsg-package.xml'
+    else:
+        fname = 'test/github-tree-data.json'
```
mikepurvis: I might have just done this as a dict, but your approach may be more robust going forward.

test/test_manifest_providers.py
```diff
@@ -46,9 +49,46 @@ def test_git_source():
...
+        text_type = str
+    else:
+        text_type = unicode
+
+    # The urlopen() function from urllib or urllib2 takes either a string or a
+    # urllib.Request object in; determine the URL in either case.
+    if isinstance(req, text_type):
+        haystack = req
+    else:
+        haystack = req.get_full_url()
```
mikepurvis: I probably would have been lazy and just done all of this as a big try block where the py3 code runs first and then errors out into the py2 code. But what you have here is fine.

clalancette:
Excellent, thanks for the reviews.  I'm going to squash all of the fixes for the tests down into a single commit (separate from the commit to add bionic), push it back up to this branch, then merge the branch.

test/test_manifest_providers.py
```diff
@@ -46,9 +49,46 @@ def test_git_source():
...
+# example from some local files.
+def mock_get_url_contents(req):
+    import re
+    import sys
+
+    # For python3, look for the 'str' type; for Python 2, the 'unicode' type
+    if sys.version_info[0] == 3:
+        text_type = str
+    else:
+        text_type = unicode
```
dirk-thomas: Please use a check which doesn't "break" in the future e.g. with Python 4, e.g.:

```
try:
    text_type = unicode
except NameError:
    text_type = str
```

test/test_manifest_providers.py
```diff
@@ -46,9 +49,46 @@ def test_git_source():
...
+    # that has the canned package.xml data we have for genmsg.  Otherwise,
+    # return the canned JSON tree data that github would have returned.
+    # In both cases, store the file-like object as a module property so that we
+    # can properly close it during 'unmock_urlopen'.
+    if re.search('.*package.xml$', haystack) is not None:
+        fname = 'test/github-genmsg-package.xml'
+    else:
+        fname = 'test/github-tree-data.json'
+
+    with open(fname, 'r') as infp:
```
dirk-thomas: What does `infp` stand for?

test/test_manifest_providers.py
clalancette: Sure, makes sense.  #110

test/test_manifest_providers.py
clalancette: infp -> "In File Pointer".  Something of a holdover from C, but I think still makes sense.

PR #111, opened 2018-01-19
fix sanitize function to collapse empty tags

dirk-thomas:
Fixes ros-infrastructure/ros_buildfarm#497.

Currently the sanitize function doesn't remove empty text nodes. As a consequence tags which are empty but contain a text node with no content are not collapsed.

As a consequence the xml fetched and stored in the cache in the first time doesn't stay the same when it is being reloaded from the cache and parsed again. The second parse will collapse the empty tags which makes the xml not equal.


mikepurvis:
👍

Issue #116, opened 2018-02-21
Wrap source repo package xmls in an object


mikepurvis:
Right now when you call `get_source_repo_package_xmls`, you get back the dict directly from the cache, but this is sucky because in order to iterate it you have to manually remove the `_ref` key that stores the SHA of the repo at the time the cache was captured.

Rather, there should be a `SourceRepoCache` object returned that implements `__getitem__`, `__iter__`, `__len__`, `keys` to expose the XML strings, but has a separate interface for getting the git reference information.

Prompted by the confusion in this discussion: https://github.com/ros-infrastructure/rosinstall_generator/pull/44#discussion_r169417262

dirk-thomas:
Addressed by #117.

PR #117, opened 2018-02-22
Add new SourceRepositoryCache class.

mikepurvis:
Addresses #116.

Testing was by rebuilding caches of my [toy rosdistro](https://github.com/mikepurvis/rosdistro-minimal) with and without the `--include-source` option, and with and without `--preclean`.

On the consumption side, I checked that `rosinstall_generator` still does the right thing with and without `--upstream-development` and against caches which do and do not have a source cache included.

### DIFF
--- a/src/rosdistro/distribution_cache.py
+++ b/src/rosdistro/distribution_cache.py
@@ -38,6 +38,7 @@ import sys
 from . import logger
 from .distribution_file import create_distribution_file
 from .package import Package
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Beside my [cosmetic change](https://github.com/ros-infrastructure/rosdistro/pull/117/commits/9c524f9c118a86dfef3721f65001898785cabaa6) this looks good to me. I am going to merge and release this. Thank you!

PR #124, opened 2018-10-27
Add ros version from REP 153

dirk-thomas:
Implementing the change from ros-infrastructure/rep#181

### DIFF
--- a/src/rosdistro/__init__.py
+++ b/src/rosdistro/__init__.py
@@ -65,7 +65,7 @@ __version__ = '0.6.9'
 
 # index information
 
-DEFAULT_INDEX_URL = 'https://raw.githubusercontent.com/ros/rosdistro/master/index.yaml'
+DEFAULT_INDEX_URL = 'https://raw.githubusercontent.com/ros/rosdistro/master/index-v4.yaml'
 
 
 def get_index_url():
--- a/src/rosdistro/index.py
+++ b/src/rosdistro/index.py
@@ -83,7 +83,12 @@ class Index(object):
                         else:
                             assert False
                     else:
-                        assert False, 'unknown key "%s"' % key
+                        if self.version <= 3:
+                            assert False, 'unknown key "%s"' % key
+                        # ignoring unknown keys
+                        # in order to allow adding new ones in the future
+                        # without the need to level the version
+                        continue
 
                     self.distributions[distro_name][key] = []
                     value = distro_data[key]
--- a/src/rosdistro/index.py
+++ b/src/rosdistro/index.py
@@ -46,8 +46,8 @@ class Index(object):
         assert data['type'] == Index._type, "Expected file type is '%s', not '%s'" % (Index._type, data['type'])
 
         assert 'version' in data, 'Index file lacks required version information'
-        assert int(data['version']) > 1, "Unable to handle '%s' format version '%d' anymore, please update your '%s' file to version '2' or '3'" % (Index._type, int(data['version']), Index._type)
-        assert int(data['version']) in [2, 3], "Unable to handle '%s' format version '%d', please update rosdistro (e.g. on Ubuntu/Debian use: sudo apt-get update && sudo apt-get install --only-upgrade python-rosdistro)" % (Index._type, int(data['version']))
+        assert int(data['version']) > 1, "Unable to handle '%s' format version '%d' anymore, please update your '%s' file to version '2', '3', or '4'" % (Index._type, int(data['version']), Index._type)
+        assert int(data['version']) in [2, 3, 4], "Unable to handle '%s' format version '%d', please update rosdistro (e.g. on Ubuntu/Debian use: sudo apt-get update && sudo apt-get install --only-upgrade python-rosdistro)" % (Index._type, int(data['version']))
         self.version = int(data['version'])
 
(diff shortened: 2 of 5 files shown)


dirk-thomas:
@jspricke @lepalom FYI. Once this is rolled out and the `rosdistro` repo is being updated to use the new index format 4 the existing Debian package of `python-rosdistro` won't be able to read the file anymore.

I am not sure if it is feasible to just release the latest version of the package (will become 0.7) or if a minimal patch to make it continue working without fully supporting format 4 is more realistic.

jspricke:
Hey Dirk, thanks for the heads up. For the package in Debian testing, there is no problem with integrating a new version (it will be frozen 2019-Jan-12, so enough time). For the stable Ubuntu versions we could ask for an SRU with a minimal patch, after we integrated the new version.

Without reading the complete discussion, can't both format be provided side by side in different URIs? This would probably help a lot of existing users (like Magazino ;) ).

dirk-thomas:
> Without reading the complete discussion, can't both format be provided side by side in different URIs?

We could create a `index-v3.yaml` file which sticks to the current format, The only downside would be that we have to maintain both files for an undefined duration into the future.

jspricke:
I would rather go for a `index-v4.yaml` and keep the old URI and yes, I would vote for keeping the old API.

dirk-thomas:
Introducing a `index-v4.yaml` file sounds appealing since it doesn't force users to upgrade. The following steps are the things which came to my mind so far which need to happen for that:

* create the new file in the `rosdistro` repo, using version 4 and adding metadata
* update unit tests in the `rosdistro` repo to ensure both files don't diverge
* release a new minor version of `python-rosdistro` adding support for index format version 4
  * I would assume this release will also change the [default URL](https://github.com/ros-infrastructure/rosdistro/blob/df28723d0d2e12b52cd81867d03ed4c979187e3a/src/rosdistro/__init__.py#L68) to point to the new v4 index
* the [buildfarm configuration](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/f352ee53a813431f3eecf1a545efc27d61b756c7/index.yaml#L119) needs to be updated to point to the new v4 index, then all maintenance jobs need to be reconfigured once, that is only the case for ROS 2 distributions, for ROS 1 distributions the code can be updated to fallback gracefully if the `distribution_type` isn't available
* for every new ROS distro both files have to be updated in the same way (a rather small price to pay for the gained backward compatibility)

I am not sure if this list is comprehensive though? Are there any other places where a hard coded `ROSDISTRO_INDEX_URL` is being used?

jspricke:
Sounds good to me.
I'm not sure if it makes sense to switch to the new format directly, maybe only if we talk about ROS2 or add a command line switch. There seem to be a lot of users of the old format and I guess we would break a lot of stuff easily: https://github.com/search?q=rosdistro%2Fmaster%2Findex.yaml&type=Code

dirk-thomas:
> I'm not sure if it makes sense to switch to the new format directly, maybe only if we talk about ROS2 or add a command line switch.

I am not sure I understand your concern. If a user has the old version of the Python package installed it will continue to use the v3 index. If the upcoming version is installed it will default to the v4 index - but since this version of `python-rosdistro` can handle that format there shouldn't be any change in behavior (except more information being available in the `dict` about a distribution). Maybe you can clarify why you think this should be delayed?

jspricke:
A normal use case is to extend the rosdistro with your own (private) repos. Depending on how this is implemented, I assume we could break existing tools and workflows. Also, if it's not need for ROS1, I would be conservative here and see how it works out first.

dirk-thomas:
> A normal use case is to extend the rosdistro with your own (private) repos. Depending on how this is implemented, I assume we could break existing tools and workflows.

@jspricke Can you please describe a concrete workflow and how it is affected by this.

> Also, if it's not need for ROS1, I would be conservative here and see how it works out first.

The goal is merge both `rosdistro` repos before the Crystal release. So there will be only one index file.

nuclearsandwich:
> I am not sure if this list is comprehensive though? Are there any other places where a hard coded `ROSDISTRO_INDEX_URL` is being used?

The configuration for build.ros2.org lives in https://github.com/ros2/ros_buildfarm_config at the moment rather than ros-infrastructure. Other than that I can't think of anything else off the top of my head.

dirk-thomas:
I created a `index-v4.yaml` file in ros/rosdistro#19415. It does the very same content as the existing `index.yaml` file and a test ensured both don't diverge from each other. So new distros are being added to both files where only the v4 one will have the additional metadata.

Existing `python-rosdistro` installation will continue to use the `index.yaml` file. In the future they might "see" ROS 2 distros in the index file though. But the same is the case as soon as they update to a newer version of the Python package which defaults to use the v4 index. So keeping ROS 2 distros out of the v3 index would only make a difference on systems where `python-rosdistro` is never being updated.

dirk-thomas:
In 4d852ae9dafb20ce38d7b1148c1cdc701ffedbc4 I relaxed the check that unknown keys raised before. As of index v4 unknown keys are just ignored.

wjwwood (approved):

missangkanathongchai694-netizen (commented):
Only just back

PR #126, opened 2019-01-04
Version match check made less restrictive

veimox:
Lossign the regex in order to also accept versions that do not contain
the "-X" suffix. This will match both "version" and "version-X".

Our use case: we do releases without the "-X suffix" which we think is a reasonable convention.

### DIFF
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -107,7 +107,7 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
             errors.append('%s: invalid package.xml file for package "%s": %s' % (dist_name, pkg_name, e))
             continue
         # check that version numbers match (at least without deb inc)
-        if not re.match('^%s-[\dA-z~\+\.]+$' % re.escape(pkg.version), repo.version):
+        if not re.match('^%s[-\dA-z~\+\.]*$' % re.escape(pkg.version), repo.version):
             errors.append('%s: different version in package.xml (%s) for package "%s" than for the repository (%s) (after removing the debian increment)' % (dist_name, pkg.version, pkg_name, repo.version))
 
         if package_xml != old_package_xml:
--- a/src/rosdistro/distribution_cache_generator.py
+++ b/src/rosdistro/distribution_cache_generator.py
@@ -107,7 +107,7 @@ def generate_distribution_cache(index, dist_name, preclean=False, ignore_local=F
             errors.append('%s: invalid package.xml file for package "%s": %s' % (dist_name, pkg_name, e))
             continue
         # check that version numbers match (at least without deb inc)
-        if not re.match('^%s[-\dA-z~\+\.]*$' % re.escape(pkg.version), repo.version):
+        if not re.match('^%s(-[\dA-z~\+\.]+)?$' % re.escape(pkg.version), repo.version):
             errors.append('%s: different version in package.xml (%s) for package "%s" than for the repository (%s) (after removing the debian increment)' % (dist_name, pkg.version, pkg_name, repo.version))
 
         if package_xml != old_package_xml:
--- a/src/rosdistro/release_cache_generator.py
+++ b/src/rosdistro/release_cache_generator.py
@@ -96,7 +96,7 @@ def generate_release_cache(index, dist_name, preclean=False, debug=False):
             errors.append('%s: invalid package.xml file for package "%s"' % (dist_name, pkg_name))
             continue
         # check that version numbers match (at least without deb inc)
-        if not re.match('^%s-[\dA-z~\+\.]+$' % re.escape(pkg.version), repo.version):
+        if not re.match('^%s(-[\dA-z~\+\.]+)?$' % re.escape(pkg.version), repo.version):
             errors.append('%s: different version in package.xml (%s) for package "%s" than for the repository (%s) (after removing the debian increment)' % (dist_name, pkg.version, pkg_name, repo.version))
 
     if not debug:


dirk-thomas:
Please provide context why you think this should be changed.

veimox:
Sure :)

Our deployment model for the robots is in three stages: testing, staging, production. We want to do deployment with a custom version of ROS where we select every single version of each package. Instead of deploying an individual package for each, we want to create a "super bundle" that installs all packages. Pretty much what [Mike Purvis presented in 2016](https://vimeo.com/187705228). 

In order to do so, a rosintall with all desired packages is created and from there, the packages are downloaded, compiled and bundled. The need comes from the creation of the rosinstall. The tool _rosinstall\_generator_ does perfectly that job and it relies on the **cached** distribution defined by a rosdistro repo. This repo can be forked, so we can add our custom pretty-much-private packages. As you can see there is no need for creating a release repo (as imposed by the standard ros buildfarm) because in our case we just need to checkout an specific tag of the source repository. 

Problem: we do releases with the _catkin\_prepare\_release_ command, which does a perfect job, but a version is tagged with the form "X.Y.Z" not "X.Y.Z-A". The second comes when is bloomed into the release repo that for this use case is not needed.

The way of defining a release of a ros package is by creating a tag following "X.Y.Z" and as such, this tool should provide the functionality of creating a cached distribution following such format and, as an extension, "X.Y.Z-A". This PR does exactly that :smile: .

dirk-thomas:
The `repo.version` in the regular expression is the version of a release entry in the rosdistro. By definition that version must have a Debian increment. What content are you placing into the `release` entry for your custom repos?

veimox:
The package file looks like this:
```yml
my_package:
  doc:
    type: git
    url: git@alotofgit.org:my-org/my-package.git
    version: upstream
  release:
    tags:
      release: {version}
    url: git@alotofgit.org:my-org/my-package.git
    version: 0.7.0
  source:
    type: git
    url: git@alotofgit.org:my-org/my-package.git
    version: kinetic-devel
  status: maintained
```
The only difference with the current system we have is that we do not use the jetkins-based buildfarm. We use Bitbucket pipelines where besides running CI, it also can create the binary for our distro. Because we generate the binary there and we only support one distro of ROS and OS, we are not in need of creating the release repository. In our source repository, we define versions with tags, and this follows the form "X.Y.Z" (in the example 0.7.0) and not "X.Y.Z-A" (e.g. 0.7.0-0). 

We want to create a .rosinstall file with rosinstall_generator with the released repos, not upstream, and to do so, we have to create our own custom cache file. This cache file is created by pulling the version from the package.xml and checking for some errors, among those, the format checking, where I am getting an error (thus the PR).

After reading the REP, it specifies:
```
version: version number for which packages are released. This can be a tag, branch or hash.
```
I cannot find where the format of the version is standarized.

Just to make sure I have checked the debian policies and [it is stated](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version) that it must follow the form `[epoch:]upstream_version[-debian_revision]`, being `epoch` and `-debian_revision` optional.

As of today, this package checks for the release "tag, branch or hash" to comply with the regex "X.Y.Z-A" being the "-A" mandatory.

PD: I have realized that this might also be a problem when defining versions with hashes, but it might be another discussion.

mikepurvis:
Not sure I'm completely following what you do here, but just to explain how we/Clearpath handle some of this stuff on our side: we don't use the `release` stanzas at all— we do everything using the `version` field of the `source` stanza, where the master branch of our internal rosdistro is the current development (devel branches) and we have forks of rosdistro for supporting previous major releases, with each point release on those being a tag of rosdistro, where each tagged rosdistro has the `version` field set to a tag or hash.

Regarding your manually-generated cache— you may not need to be doing that. Note that it's not well advertised/documented, but have a play with the source repo cache generated by `rosdistro_build_cache --include-source` and note that `rosinstall_generator --upstream-development` will use it, if available. For a variety of (legitimate) reasons, the public buildfarm doesn't generate the source cache by default, which is part of why discoverability for it is so poor. :(

Anyway, all that said, we also haven't found a great model for managing ongoing forks with upstream, and loosening the strict version number format would definitely give us the flexibility to use suffixes (eg ros_comm 1.14.3cpr1 -> 1.14.3cpr2 -> 1.14.4 -> 1.14.4cpr1), which is similar to what Debian/Ubuntu does. As it stands, we mostly just generate arbitrary tags (see my [simple patching/tagging tool](https://github.com/mikepurvis/forkyeah#forkyeah)), and the package.xml version remains unchanged from what it is from the last upstream version. This renders it mostly meaningless— for a forked repo, you have to go back to the rosdistro tag to see what actual version of the source + patches ended up in there.

veimox:
@dirk-thomas I have updated the regex and set in both places.

@mikepurvis I see two parts in the comment: one is related to the PR, the suffixes; and the other it might be to be discussed in another forum,  other deployment models than the buildfarm. If you are interested we can open a thread in discourse (Quality-Assurance? Buildfarm?) :)

Regarding the suffixes, as proposed, it breaks the debian standard (does not contain the -), the last commit imposes this character in case a patch exists so it would look like "1.14.3-cpr1 -> 1.14.3-cpr2 -> 1.14.4 -> 1.14.4-cpr1". I am not sure how can this affect your use case.

Regarding the prefixes, maybe the regex should also include the [epoch:], even though is not being used I think is more important follow the debian guideline. I proppose `^([\dA-z]+?:)%s(-[\dA-z~\+\.]+)?$` [(regex101)](https://regex101.com/r/4ZkE53/1)

Finally, regarding the deployment model discussion. Making it short, we are following a model in which, as a general rule, we do the next:

1. System packages: we rely fully on the OS to select (e.g. version of gcc).
2. OSRF packages: we rely on the APT repo for the previous distro.
3. Company packages: we use our own CI to binarize and deploy into our own APT repo.

That being PER package.

Now we want to create a "super bundle" as you proposed back in 2016 but this has to comply with the guidelines above. In `rosinstall_generator` the flag `--upstream-development` can be used but that would imply using the source repositories. In a sense, that is relying on another source of information that, ideally is the same (the tag 0.1.0 in the source repo should contain the same code as the release repo in 0.1.0) but as it is human modifiable, it might not be. Besides, there are repos that contain the release repository but not the source. Because of this need of using the release track, I (think) am in need of generating a custom cache, but that conflicts we the releasing system we currently follow (thus the PR).

I would love to discuss more in depth and if you (plural) are interested in discussing this and a more production oriented topic, we can open a thread in discourse to start with. There are other methods as Docker, Valgrant, Snaps available and we don't have the perfect solution either :smile:

mikepurvis:
My reasoning about the dash (and in particular not including it) is that the dash separates the _source/upstream version_ from the _packager's version_. As things are set up today, the `package.xml` file is supposed to contain the source/upstream version, hence why it's so strict in how you're supposed to format that number— by not permitting dashes or colons or whatever, you leave all that extra stuff to be used at packaging time.

If we're maintaining a version of source that's forked from what is called `1.14.3` by upstream, IMO what we're maintaining is still _source_, so I'd still rather not include the dash. That is, I'd prefer to call it `1.14.3cpr1` and not `1.14.3-cpr1`. Consider the case where such a fork was bloomed and released as a buildfarm deb— the version of that deb should be `1.14.3cpr1-0`, maintaining the idea that `package.xml`'s version field is the source version.

tfoote:
I agree on holding the distinction between dashes being separating the packaging level version to allow us to be inherently compatible with the debian versioning: https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

Related fedora only allows ascii in the right most of the triplet in addition to the whole numbers.
https://fedoraproject.org/wiki/Packaging:Versioning#Simple_versioning

And note that '~' breaks RPMs so that's probably good to keep off the table. It's also more commonly used in the debian context in the debian increment section not the upstream version number. They don't seem to call that out

dirk-thomas:
The patch looks good now and will enable the described use case (referencing a tagged version of the source repo in the release entry). Therefore I am inclined to merge this if nobody wants to argue against it.

---

The `ros/rosdistro` will continue to enforce a Debian increment in the release sections since it requires the entry to point to a release repository containing Debian control files etc.

PR #129, opened 2019-01-29
bump Python 3 minimum version to 3.4

dirk-thomas:
Follow up of #123.


dirk-thomas:
Worked around broken GitHub resources by changing the tests to use a different tag: #130.

PR #132, opened 2019-03-19
Fix rosdistro_freeze_source for Python3

mikepurvis:
Validated by freezing a toy rosdistro repo and using the same source under py2 and py3 venvs.

FYI @tspicer01

### DIFF
--- a/src/rosdistro/freeze_source.py
+++ b/src/rosdistro/freeze_source.py
@@ -51,7 +51,7 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
     # Populate this queue with tuples of repositories instances to be updated,
     # so that this work can be spread across multiple threads.
     work_queue = queue.Queue()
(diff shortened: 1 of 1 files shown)


src/rosdistro/freeze_source.py
```diff
@@ -84,7 +84,7 @@ def freeze_distribution_sources(dist, release_version=False, release_tag=False,
 def _get_repo_info(url, retry=2, retry_period=1):
     cmd = ['git', 'ls-remote', url]
     try:
-        return subprocess.check_output(cmd).splitlines()
+        return subprocess.check_output(cmd).decode().splitlines()
```
mikepurvis: The returned value here was previously a list of strings. Now it is on Python 2 a list of unicodes, and on Python 3 a list of str.

dirk-thomas:
Thanks for the patch.

dirk-thomas:
Do you need a new release for this fix?

PR #133, opened 2019-03-20
github.py:  encode('utf-8') package XML before parsing

zultron:
The `_get_url_contents()` function decodes downloaded content before
returning it.

The `github_source_manifest_provider()` function calls
`package.parse_package_string()` on this content, which passes it to
`xml.dom.minidom.parseString()`.  That expects input to be encoded.

This patch encodes that content before sending it out to
`package.parse_package_string()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ deb_dist
 dist
 *.pyc
 rosdistro.egg-info
+
+# nosetests artifacts
+/.coverage
+/nosetests.xml
--- a/src/rosdistro/manifest_provider/github.py
+++ b/src/rosdistro/manifest_provider/github.py
@@ -119,7 +119,7 @@ def github_source_manifest_provider(repo):
             (path, cache.ref(), package_xml_path + '/package.xml' if package_xml_path else 'package.xml')
         logger.debug('- load package.xml from %s' % url)
         package_xml = _get_url_contents(url)
-        name = parse_package_string(package_xml).name
+        name = parse_package_string(package_xml.encode('utf-8')).name
         cache.add(name, package_xml_path, package_xml)
 
     return cache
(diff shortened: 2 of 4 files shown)


zultron:
Earlier I erroneously filed a PR at ros-infrastructure/catkin_pkg#252.  Clearly I don't understand my python Unicode, so please check this carefully.

dirk-thomas:
There are numerous manifest files with special characters in already released packages for which the process seem to work just fine. Please provide more information how to reproduce the problem.

zultron:
@dirk-thomas Thanks for looking at this.  I can't explain why this bug isn't triggered in released packages, but I have reproduced the bug in the tests, narrowed it down to Python 2 only, and with some incremental pushing the results are visible in the Travis tests in this PR.

The second-to-top commit adjusts the tests to better mimic the behavior of `rosdistro.manifest_provider.github._get_url_contents()` by decoding the XML bytes into a Unicode string before returning it.  The bug is then visible in the Python 2 tests, shown [failing in this test run][tci-fail].

The top commit is the fix in the original PR, shown [passing in this run][tci-pass].

I put as much explanation as I'm able into the commit log, but I'm not able to explain what changed in Python 3 (or how this bug escaped attention in the past).

[tci-fail]: https://travis-ci.org/ros-infrastructure/rosdistro/builds/509268781
[tci-pass]: https://travis-ci.org/ros-infrastructure/rosdistro/builds/509272184

dirk-thomas:
I don't think patching the parameter of `parse_package_string()` is the right approach. Please see ros-infrastructure/catkin_pkg#253 for an alternative fix making the function more flexible.

zultron:
I'm fine with however you want to fix this.  The first commit that fixes the unit tests still ought to stay, IMO, even if you ultimately go with the fix in ros-infrastructure/catkin_pkg#253.  Let me know and I can pop the top commit off this PR.

zultron:
...upon second look at the other PR, I think the unit test change here can go, too, since it's incomplete and doesn't test both cases.  I'm closing this one out.

PR #134, opened 2019-03-22
Sync dep_type with catkin_pkg/package.py to allow more query types

andre-rosa:
* https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L273
  * To be used by superflore as in https://discourse.ros.org/t/a-proposal-for-a-superflore-oe-recipe-generation-scheme/8401

### DIFF
--- a/src/rosdistro/dependency_walker.py
+++ b/src/rosdistro/dependency_walker.py
@@ -118,10 +118,14 @@ class DependencyWalker(object):
     def _get_dependencies(self, pkg_name, dep_type):
         pkg = self._get_package(pkg_name)
         deps = {
-            'buildtool': pkg.buildtool_depends,
             'build': pkg.build_depends,
+            'buildtool': pkg.buildtool_depends,
+            'build_export': pkg.build_export_depends,
+            'buildtool_export': pkg.buildtool_export_depends,
+            'exec': pkg.exec_depends,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please provide some context why you need the different kind of dependencies? The differentiation of `*tool` dependencies shouldn't be relevant in non cross-compilation scenarios and the different build and run dependencies will likely be treated in the same way, no?

andre-rosa:
@dirk-thomas << We're working on [extending superflore](https://discourse.ros.org/t/a-proposal-for-a-superflore-oe-recipe-generation-scheme/8401) to automatically generate buildable Bitbake recipes of ROS2 index packages, focusing on the latest ROS 2 and OpenEmbedded/Yocto releases.

For that cross-compilation scenario, we need to be able to differentiate between a native package (runs on the host) and non-native package (runs on the target), as described in [proposal](https://github.com/lgsvl/superflore/wiki/Proposed-Superflore-OE-Recipe-Generation-Scheme).

We use *tool dependencies to indicate '-native' differentiating it from non-native.

dirk-thomas:
Sounds good. Since the function only returns information for the given `dep_type` this shouldn't affect existing code.

andre-rosa:
Agree; thank you for reviewing!

PR #135, opened 2019-03-29
Factor out an independent function to fetch a distribution cache file

andre-rosa:
* No functional change and no side effects
  * Decrease file manipulation scope by closing it just after reading
  * Will be used by OpenEmbedded support in superflore

### DIFF
--- a/src/rosdistro/__init__.py
+++ b/src/rosdistro/__init__.py
@@ -149,7 +149,7 @@ def get_cached_distribution(index, dist_name, cache=None, allow_lazy_load=False)
     return dist
 
 
-def get_distribution_cache(index, dist_name):
+def get_distribution_cache_string(index, dist_name):
     if dist_name not in index.distributions.keys():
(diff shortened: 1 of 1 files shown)


src/rosdistro/__init__.py
```diff
@@ -149,7 +149,7 @@ def get_cached_distribution(index, dist_name, cache=None, allow_lazy_load=False)
     return dist
 
 
-def get_distribution_cache(index, dist_name):
+def fetch_distribution_cache_file(index, dist_name):
```
dirk-thomas: Since the function doesn't fetch just the file (which might be gzipped) how about renaming the new function to `get_distribution_cache_string`?

andre-rosa: Good idea, thank you!

dirk-thomas:
Thank you for the improvement.

andre-rosa:
Wow, thank _you_ @dirk-thomas for being so diligent!

PR #136, opened 2019-04-23
Allow evaluating dependency conditions in dependency walker

andre-rosa:
- Add the optional capability to evaluate dependency conditions
  - It'll be used in superflore to avoid pulling in ROS 1 dependencies
    for packages that list dependencies conditionally on ROS_VERSION
    like ml-classifiers and we're building ROS 2.
    https://github.com/astuff/ml_classifiers/blob/master/package.xml
    ...
    <build_depend condition="$ROS_VERSION == 1">message_generation</build_depend>
    <build_depend condition="$ROS_VERSION == 2">rosidl_default_generators</build_depend>
    ...

### DIFF
--- a/src/rosdistro/dependency_walker.py
+++ b/src/rosdistro/dependency_walker.py
@@ -36,9 +36,10 @@ from catkin_pkg.package import InvalidPackage, parse_package_string
 
 class DependencyWalker(object):
 
-    def __init__(self, distribution_instance):
+    def __init__(self, distribution_instance, evaluate_condition_context=None):
         self._distribution_instance = distribution_instance
         self._packages = {}
+        self._evaluate_condition_context = evaluate_condition_context
 
     def _get_package(self, pkg_name):
         if pkg_name not in self._packages:
@@ -50,6 +51,9 @@ class DependencyWalker(object):
                 pkg = parse_package_string(pkg_xml)
             except InvalidPackage as e:
                 raise InvalidPackage(pkg_name + ': %s' % str(e))
+            else:
+                if self._evaluate_condition_context is not None:
+                    pkg.evaluate_conditions(self._evaluate_condition_context)
             self._packages[pkg_name] = pkg
         return self._packages[pkg_name]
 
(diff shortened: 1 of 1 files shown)


src/rosdistro/dependency_walker.py
```diff
@@ -48,7 +49,9 @@ def _get_package(self, pkg_name):
             pkg_xml = self._distribution_instance.get_release_package_xml(pkg_name)
             try:
                 pkg = parse_package_string(pkg_xml)
-            except InvalidPackage as e:
+                if self._evaluate_condition_context:
```
dirk-thomas: Please check for `is not None` explicitly.

Same below.

src/rosdistro/dependency_walker.py
```diff
@@ -48,7 +49,9 @@ def _get_package(self, pkg_name):
...
+                    pkg.evaluate_conditions(self._evaluate_condition_context)
```
dirk-thomas: Please move the lines 52 and 53 into an `else` block after `except` since they can't raise an `InvalidPackage` exception. In that `else` you can use another `try` / `except` for `ValueError` but raising `InvalidPackage` doesn't seem to make sense in that case.

src/rosdistro/dependency_walker.py
```diff
@@ -127,6 +130,8 @@ def _get_dependencies(self, pkg_name, dep_type):
             'test': pkg.test_depends,
             'doc': pkg.doc_depends,
         }
+        if self._evaluate_condition_context:
+            return set([d.name for d in deps[dep_type] if d.evaluated_condition])
```
dirk-thomas: To avoid the condition you could always check for `if d.evaluated_condition is not False`.

src/rosdistro/dependency_walker.py
andre-rosa: Done, thanks!

src/rosdistro/dependency_walker.py
andre-rosa: Done!

dirk-thomas:
Thank you for the improvement.

Issue #138, opened 2019-07-12
Argument number does not match


danielwangksu:
`_yaml_header_lines` requires 2 argument
https://github.com/ros-infrastructure/rosdistro/blob/2565c2c5e4add200dc8828d738e7c3d2c1850485/scripts/rosdistro_migrate_to_rep_141#L152

https://github.com/ros-infrastructure/rosdistro/blob/2565c2c5e4add200dc8828d738e7c3d2c1850485/src/rosdistro/verify.py#L160

dirk-thomas:
Please consider to provide a pull request to address the problem.

PR #141, opened 2019-07-18
fixing argument number mismatch

danielwangksu:
fix https://github.com/ros-infrastructure/rosdistro/issues/138
Signed-off-by: Daniel Wang <daniel.wang@canonical.com>


dirk-thomas:
Thank you for the fix.

Issue #139, opened 2019-07-12
Modification of parameter with default


danielwangksu:
Python’s default arguments are evaluated once when the function is defined, not each time the function is called. This might cause unexpected behavior
https://github.com/ros-infrastructure/rosdistro/blob/2565c2c5e4add200dc8828d738e7c3d2c1850485/src/rosdistro/aptdistro.py#L32-L36

https://github.com/ros-infrastructure/rosdistro/blob/2565c2c5e4add200dc8828d738e7c3d2c1850485/src/rosdistro/aptdistro.py#L44-L48

```
   def depends_on(self, package, res=None, one=False):
        for p, dep in self.dep.iteritems():
            if res is None:
               res = []
            if package in dep:
                if p not in res:
                    res.append(p)
                if not one:
                    self.depends_on(p, res, one)
        return res
```

dirk-thomas:
Please consider to provide a pull request to address the problem.

PR #142, opened 2019-07-18
fixing modification of parameter with default

danielwangksu:
fix https://github.com/ros-infrastructure/rosdistro/issues/139
Signed-off-by: Daniel Wang <daniel.wang@canonical.com>


dirk-thomas:
Thanks for the fix.

PR #151, opened 2020-04-13
Support tar sources and releases

kyrofa:
Both the `distribution.yaml` and rosinstall files support non-git sources. However, rosdistro currently only supports building caches using git repositories. Add support for tarfiles as well.

### DIFF
--- a/src/rosdistro/distribution.py
+++ b/src/rosdistro/distribution.py
@@ -34,13 +34,14 @@
 from .manifest_provider.bitbucket import bitbucket_manifest_provider
 from .manifest_provider.git import git_manifest_provider, git_source_manifest_provider
 from .manifest_provider.github import github_manifest_provider, github_source_manifest_provider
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Please check the failing CI. Two cases will be addressed by releasing a new version of `catkin_pkg` with ros-infrastructure/catkin_pkg#282 but the other cases seem to be caused by this patch.

kyrofa:
Whoops, sorry about that. Should be all fixed now, other than the two known issues.

dirk-thomas:
`catkin_pkg` 0.4.17 has been released but the Travis config in this repo needed some poking too, see #152. Please rebase one more time.

kyrofa:
Of course, no trouble. Done and green. Thanks for taking care of those!

dirk-thomas:
Thanks for the enhancement.

PR #156, opened 2020-08-30
[Windows] More portable way to check Git executable

seanyen:
On Windows, by its convention, the executable has a file extension `.exe` and this pull request is to add that for searching Git executable.

### DIFF
--- a/src/rosdistro/vcs.py
+++ b/src/rosdistro/vcs.py
@@ -46,8 +46,7 @@ class Git(object):
     def __init__(self, cwd=None):
         self.cwd = cwd
         if not self._client_executable:
-            git_excutable = 'git' if os.name != 'nt' else 'git.exe'
-            self.__class__._client_executable = _find_executable(git_excutable)
+            self.__class__._client_executable = _find_executable('git')
 
     def command(self, *args):
         assert self._client_executable is not None, "'git' not found"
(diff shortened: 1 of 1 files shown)


cottsay:
Will this handle `git.cmd` too?

mikepurvis:
I would handle this kind of thing in the `_find_executable` helper itself, though I suppose in this case it's private to the module so it doesn't much matter:

https://github.com/ros-infrastructure/rosdistro/blob/15198de618b6fcd1dd6160fc676b9c544d102a15/src/rosdistro/vcs.py#L83-L88

dirk-thomas:
> I would handle this kind of thing in the `_find_executable` helper itself

I don't think appending `.exe` is in 100% of the cases the correct assumption.

seanyen:
> Will this handle `git.cmd` too?

Where was this `git.cmd`? Is it a wrapper for Git? I am using `git version 2.27.0.windows.1` and I didn't see this file.

cottsay:
> Where was this `git.cmd`?

I'm not aware of any setups which install `git.cmd` by default, but while I was at Microsoft, we used `*.cmd` wrappers all the time to avoid adding excess directories to our `PATH`.

Certainly not a blocker, but I was curious. I would think that a complete implementation of `_find_executable` would be able to find `*.cmd` files as well as `*.exe`, since that's the behavior exhibited by `where` and CMD itself.

seanyen:
> Certainly not a blocker, but I was curious. I would think that a complete implementation of `_find_executable` would be able to find `*.cmd` files as well as `*.exe`, since that's the behavior exhibited by `where` and CMD itself.

Ah I see. In such case, I think we can peak `%PATHEXT%` to append the possible file extensions which `CMD.exe` recognize. [Link](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start#remarks).

dirk-thomas:
> I think we can peak `%PATHEXT%` to append the possible file extensions which `CMD.exe` recognize.

That sounds good to me. We already do the same in e.g. `ros2 run`. Can you please update the PR accordingly?

dirk-thomas:
Thanks for the patch.

PR #157, opened 2020-09-02
[Windows] Adding a helper to deal with read-only files

seanyen:
On Windows, some `.git` metadata files are marked as read-only by default, and in this case, the `rmtree` will result in the following error:

```
DEBUG:rosdistro:Skipped "github_manifest_provider()": VCS url "https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs-release.git" does not match expected format.
DEBUG:rosdistro:Skipped "bitbucket_manifest_provider()": VCS url "https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs-release.git" does not match expected format.
DEBUG:rosdistro:Skipped "git_manifest_provider()": Unable to fetch package.xml: [WinError 5] Access is denied: 'C:\\Users\\tzuhs\\AppData\\Local\\Temp\\tmp3f0d2xelrosdistro\\.git\\objects\\05\\10a7e44d448270128a8c99c08d2808e647803c'
```

And eventually `rosdistro` won't be able to see and load the `package.xml` file and show a similar message like below:

```
dashing: missing package.xml file for package "autoware_auto_msgs"
```

This pull request is to add a error handling to `rmtree`, attempt to modify the file attributes, and retry again.


src/rosdistro/manifest_provider/git.py
```diff
@@ -107,4 +116,4 @@ def _temp_git_clone(url, ref):
 
         yield base
     finally:
-        shutil.rmtree(base)
+        shutil.rmtree(base, onerror=onerror)
```
dirk-thomas, line 119: The `onerror` handler is only necessary on Windows and therefore should only be added on that platform.

src/rosdistro/manifest_provider/git.py
```diff
@@ -83,6 +83,15 @@ def git_source_manifest_provider(repo):
     return cache
 
 
+def onerror(func, path, exc_info):
+    import stat
+    if not os.access(path, os.W_OK):
```
dirk-thomas, line 88: Instead of checking the access permissions how about checking the exception?

src/rosdistro/manifest_provider/git.py
```diff
@@ -83,6 +83,15 @@ def git_source_manifest_provider(repo):
     return cache
 
 
+def onerror(func, path, exc_info):
+    import stat
```
dirk-thomas, line 87: This should be imported on the module level if there is no reason not to do so.

src/rosdistro/manifest_provider/git.py
```diff
@@ -83,6 +83,15 @@ def git_source_manifest_provider(repo):
...
+    if not os.access(path, os.W_OK):
+        os.chmod(path, stat.S_IWUSR)
+        func(path)
+    else:
+        raise
```
dirk-thomas, line 92: Is there an active exception which can be reraised? Have you tried this code path?

dirk-thomas (commented):
Overall an approach like this https://github.com/dirk-thomas/vcstool/blob/ff8ce2638a17a80e3ee303db502b27e1d7dabdcf/vcstool/util.py#L9-L18 might be better.

seanyen:
> Overall an approach like this https://github.com/dirk-thomas/vcstool/blob/ff8ce2638a17a80e3ee303db502b27e1d7dabdcf/vcstool/util.py#L9-L18 might be better.

That's a good reference! I am shamelessly stealing the logic and apply it to all the `rmtree` usages in `rosdistro`.

src/rosdistro/util.py
```diff
@@ -0,0 +1,51 @@
+# Software License Agreement (BSD License)
+#
+# Copyright (c) Microsoft Corporation.
```
dirk-thomas, line 3: Since you copy-n-pasted the code snippet I shared from `vcstool` claiming copyright is certainly not right.

PR #159, opened 2020-09-18
rework the UTF-8 fix for Python2/3 compatibility.

seanyen:
This is a rework to my previous commit https://github.com/ros-infrastructure/rosdistro/pull/158.

The original fix using `encoding` argument is not Python2 friendly. It is a feature only existing on Python3.


dirk-thomas:
Thanks for the fix.

PR #163, opened 2021-10-16
Add a GitLab manifest provider

cottsay:
Needs testing and authentication support


cottsay:
@mikepurvis, are you still interested in GitLab support here? Any chance you can give me a hand in testing this and updating it to work with on-prem URLs?

mikepurvis:
@cottsay Sorry I missed this tag. I no longer needed this as we implement our own asyncio version of this kind of thing, see:

https://github.com/clearpathrobotics/colcon-distro/blob/main/colcon_distro/download.py#L139-L155

But hopefully it can be useful to someone else.

src/rosdistro/manifest_provider/gitlab.py
```diff
@@ -0,0 +1,156 @@
...
+    if GITLAB_PRIVATE_TOKEN:
+        req.add_header('Private-Token', GITLAB_PRIVATE_TOKEN)
+    logger.warn('Performing GitLab API query "%s"' % (url,))
+    return urlopen(req)
+
+
+def _gitlab_api_query(server, path, resource, attrs):
+    url = 'https://%s/api/v4/projects/%s/%s' % (server, urlquote(path, safe=''), resource)
+    if attrs:
+        url += '?' + '&'.join(k + '=' + urlquote(str(v), safe='') for k, v in attrs.items())
```
nuclearsandwich, line 59: Is there a reason not to use [`urllib.parse.urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)?

```suggestion
        url += '?' urlencode(attrs)
```

src/rosdistro/manifest_provider/gitlab.py
```diff
@@ -0,0 +1,156 @@
...
+            match = re.match(r'.*<([^>]*)>; rel="next"', links)
+            if not match:
+                break
+            url = match.group(1)
+
+
+def gitlab_manifest_provider(_dist_name, repo, pkg_name):
+    assert repo.version
+    server, path = repo.get_url_parts()
+    if not server.startswith('gitlab.'):
```
nuclearsandwich, line 93: I think it's a minor concern, but there are GitLab instances which don't have gitlab in the domain, e.g. <https://salsa.debian.org>. I think some kind of override or indicator for a GitLab handling via environment variable or YAML type sigil could work as a future enhancement but it's not worth blocking 2 on.

nuclearsandwich (approved):
One question and one suggestion that possibly simplifies things. Thanks for getting this one across the finish line.

src/rosdistro/manifest_provider/gitlab.py
cottsay: No reason at all. 1b46c7b

src/rosdistro/manifest_provider/gitlab.py
cottsay: I'm also not totally sold on this. Maybe we take this as-is and add an escape hatch later if needed...?

mikepurvis: Escape hatch could be just an environment variable like `ROSDISTRO_GITLAB_SERVER` that if set has to be an exact match. That allows anyone with a private instance on a funky domain to set it, and any public ones (like salsa) can just be added here as needed.

cottsay: Alright, I added it as suggested, but I also removed the hacky `server.startswith('gitlab.')` in favor of the pattern already used for GitHub, `server.endswith('gitlab.com')` since we don't need that heuristic anymore.

nuclearsandwich (approved):

PR #167, opened 2022-03-07
Run tests with pytest instead of nose

cottsay:
Official nose documentation recommends users migrate to a more supported platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ deb_dist
 dist
 *.pyc
 rosdistro.egg-info
+.coverage
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ testsetup:
 	echo "running rosdistro! tests"
 
 test: testsetup
-	python setup.py nosetests
+	cd test && pytest
 
 test--pdb-failures: testsetup
-	python setup.py nosetests --pdb-failures
+	cd test && pytest --pdb
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,0 +1,6 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.join(
+    os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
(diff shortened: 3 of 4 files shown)


test/__init__.py
```diff
@@ -0,0 +1,6 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.join(
+    os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
+    'src'))
```
cottsay, line 6: Nose seems to be modifying the path so that the module in the repository directory is prefixed on the path, and pytest doesn't seem to do that (at least when a `src` subdirectory is used).

This is the approach that `catkin_pkg` uses, and it seems to work fine here as well.

.gitignore
```diff
@@ -4,3 +4,4 @@ deb_dist
 dist
 *.pyc
 rosdistro.egg-info
+.coverage
```
nuclearsandwich, line 7: I prefer the .gitignore lines to be lexicographically sorted (and prefer directories to have trailing slashes) but since neither of those conventions are adhered to in the current contents moving .coverage to the top is optional as I'll have to make a separate PR bringing us into conformance generally.

Makefile
```diff
@@ -29,7 +29,7 @@ testsetup:
 	echo "running rosdistro! tests"
 
 test: testsetup
-	python setup.py nosetests
+	cd test && pytest
```
nuclearsandwich, line 32: Why is `cd test` needed where it wasn't previously?

nuclearsandwich (approved):
One non-critical nitpick and one question aside, this looks good!

Makefile
cottsay: I did that to conform with the pattern used in other `ros-infrastructure` packages.

.gitignore
cottsay: No reason to make things worse. Done in e03cc3d.

PR #168, opened 2022-03-07
List 'ROS Infrastructure Team' as the package maintainer

cottsay:
This package is currently maintained by members of the ROS infrastructure team at Open Robotics.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,12 @@ kwargs = {
     'author_email': 'wim@hidof.com, dthomas@osrfoundation.org',
     'maintainer': 'ROS Infrastructure Team',
     'url': 'http://wiki.ros.org/rosdistro',
+    'project_urls': {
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@cottsay, while I don't _want_ people to email maintainers, whoever they are, I do want to make sure that we provide some kind of contact mechanism and so I was a bit hesitant to remove the maintainer_email field without any replacement.

I pushed 74e8315 which adds [project_urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) for the source code and issue tracker. If you think that's a welcome change I'll start updating the other packages as well. Eventually we should consider both updating the `url` "homepage" and making sure the documentation for the ros-infrastructure packages is in good shape before including documentation links.

nuclearsandwich (approved):

PR #174, opened 2024-03-08
Switch to a raw string for various regexes.

clalancette:
Otherwise, when running against Python 3.12, we see warnings like SyntaxWarning: invalid escape sequence '\d'

This should fix warnings that we see in jobs like https://build.ros2.org/view/Queue/job/Rsrc_uN__ackermann_msgs__ubuntu_noble__source/1/consoleFull


cottsay (approved):
This looks good once we get CI fixed.

PR #176, opened 2024-04-04
Switch to ros-infrastructure centralized CI

cottsay:
Closes #175

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,33 +1,21 @@
-name: rosdistro-ci
+---
+name: Run tests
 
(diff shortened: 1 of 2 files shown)


cottsay:
Oh lame, we're actually writing out yaml files that don't meet the standard `yamllint` configuration. I'll have to garner the mental energy to deal with this some other time.

cottsay:
> Oh lame, we're actually writing out yaml files that don't meet the standard `yamllint` configuration.

I specifically skipped these files from the linter. Still lame though.

nuclearsandwich (approved):

PR #182, opened 2024-05-28
Add get_package_condition_context function

cottsay:
This function captures package condition context variables for use when resolving conditions in package manifests. This function is currently [implemented in ros_buildfarm](https://github.com/ros-infrastructure/ros_buildfarm/blob/340010179c047aa319416a9be74ef89487841688/ros_buildfarm/common.py#L743), but is useful in other contexts as well.

### DIFF
--- a/src/rosdistro/__init__.py
+++ b/src/rosdistro/__init__.py
@@ -179,6 +179,29 @@ def get_distribution_cache(index, dist_name):
     return DistributionCache(dist_name, data)
 
 
+def get_package_condition_context(index, dist_name):
+    if dist_name not in index.distributions.keys():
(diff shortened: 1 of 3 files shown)


mikepurvis:
Looks good. I implemented something similar in colcon-distro to correctly handle chimera packages like plotjuggler, where the same source branches are being built for ROS 1 and 2 distributions.

cottsay:
> Looks good.

Thanks for the feedback. GitHub reviews are always welcome on the infrastructure team, even if the check mark isn't green :)

mikepurvis (approved):

nuclearsandwich (approved):

PR #184, opened 2024-06-25
Remove more Python 2 artifacts

cottsay:
We no longer support Python 2 in rosdistro so there's no reason to carry these snippets anymore.

### DIFF
--- a/scripts/rosdistro_build_cache
+++ b/scripts/rosdistro_build_cache
@@ -90,9 +90,7 @@ def main():
         if args.dist_names and dist_name not in args.dist_names:
             continue
         # write the cache
-        data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper)
-
-        data = data.encode('utf-8')
+        data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper).encode()
 
         with open('%s-cache.yaml' % dist_name, 'wb') as f:
             print('- write cache file "%s-cache.yaml"' % dist_name)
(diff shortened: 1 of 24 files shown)


scripts/rosdistro_build_cache
```diff
@@ -94,11 +92,7 @@ def main():
         # write the cache
         data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper)
 
-        # On Python 3, we must encode the unicode yaml str prior to writing it,
-        # whereas for Python 2, the yaml output is a str which cannot be further
-        # encoded (compared with a unicode object).
-        if sys.version_info[0] >= 3:
-            data = data.encode('utf-8')
+        data = data.encode('utf-8')
```
mikepurvis, line 95: data = yaml.dump(cache.get_data(), Dumper=CacheYamlDumper).encode()

cottsay: Would omitting `'utf-8'` constitute a change in behavior on any platforms?

mikepurvis: It's the default: https://docs.python.org/3/library/stdtypes.html#str.encode

mikepurvis: A better way still might be to just let the encode happen implicitly by changing `open('wb')` to just `open('w')`, as it looks like gzip.open also has a textual mode that presumable does the same thing: https://docs.python.org/3/library/gzip.html#gzip.open

But I can also understanding wanting to keep this change simple. :)

cottsay: I'll take the suggestion, but let's not rock the boat too much :) a1b4ba9

nuclearsandwich (approved):

PR #185, opened 2024-06-28
Clean up script entry points

cottsay:
Two factors here:
1. Ensure that each script has an explicit main() function
2. Ensure that the return value of main() is the return value of the program when the script is invoked directly.

### DIFF
--- a/scripts/rosdistro
+++ b/scripts/rosdistro
@@ -83,4 +83,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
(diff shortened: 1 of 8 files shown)


christophebedard (approved):
Looks reasonable to me.

I'm seeing a bit of a mix of single and double quotes, but that's probably outside of the scope of this PR.

cottsay:
> I'm seeing a bit of a mix of single and double quotes, but that's probably outside of the scope of this PR.

There's actually no `flake8` check on this repo at all. There are a LOT of deviations and inconsistencies...

PR #189, opened 2024-06-28
Remove unused README.rst

cottsay:
There is also a README.md file in this repository, and that's what visitors see on GitHub.


nuclearsandwich (approved):
The setup.py long description is a verbatim string so even pypi is not using this file https://github.com/ros-infrastructure/rosdistro/blob/master/setup.py#L48

PR #191, opened 2024-10-11
Fix entry_points removal for SKIP_PYTHON_SCRIPTS

cottsay:
I needed this workaround to publish 1.0.0. Setuptools got mad about it.

Follow-up to 8de70e46840c65383d5a0791cade5b932ffe999c


nuclearsandwich (commented):
Retroactive ✅.

PR #194, opened 2024-12-11
Drop distutils as it is deprecated

MatthijsBurgh:


MatthijsBurgh:
@tfoote @cottsay @mikepurvis please review this PR.

tfoote (changes requested):
Sorry for the long delay reviewing. Thanks for contributing this. It's luckily a small tweak. 

Can you rebase to a more recent commit to get the CI runs to trigger? Otherwise this looks great to add.

MatthijsBurgh:
@tfoote rebase done

I need approval for a workflow run

MatthijsBurgh:
The version of the git client on windows is `2.51.0.windows.2`. It is not accepted by packaging. We either have to do pre-processing. Or keep using the `distutils` module shipped by `setuptools`. @cottsay what do you think?

MatthijsBurgh:
@cottsay please respond to my previous comment

cottsay:
I don't have any silver bullets here. We clearly need to get away from distutils, so I'd guess we'll have to do some intermediate processing, which is a bummer. I'm open to suggestions.

MatthijsBurgh:
@cottsay @tfoote please approve workflow run

MatthijsBurgh:
I have no clue why the CI fails. As it runs on my account, https://github.com/MatthijsBurgh/rosdistro-1/actions/runs/21364087466

I had to restart it a few times. It looks more like a bug in GH actions than a failing pipeline.

cottsay:
> I had to restart it a few times. It looks more like a bug in GH actions than a failing pipeline.

Agreed. Looks like they're finally acknowledging that there's an issue: https://www.githubstatus.com/incidents/90hj03y5tj3c

cottsay:
This will need an update to `setup.py` and `stdeb.cfg` to include the new dependency.

MatthijsBurgh:
> This will need an update to `setup.py` and `stdeb.cfg` to include the new dependency.

@cottsay Done

MatthijsBurgh:
@cottsay please approve workflow run and review

mikepurvis (approved):

setup.py
```diff
@@ -10,7 +10,7 @@
     # - src/rosdistro/__init__.py
     # - stdeb.cfg
     'version': '1.0.1',
-    'install_requires': ['PyYAML', 'setuptools'],
+    'install_requires': ['packaging>=17', 'PyYAML', 'setuptools'],
```
cottsay: Is there any particular reason this is bound to version 17? We have 16.8 in RHEL 8, which appears to work.

MatthijsBurgh: https://github.com/colcon/colcon-cmake/pull/156#discussion_r2728778851
But we don't use that feature here.

But packaging 22, adds https://github.com/pypa/packaging/commit/237ff3aa348486cf835a980592af3a59fccd6101. Before this commit, `parse` will return a `LegacyVersion`, in case of an `InvalidVersion`. These compare not correctly.

```py
LegacyVersion('0.1.0.windows.2') >  Version('0.2.0')=False
LegacyVersion('0.1.0.windows.2') >= Version('0.2.0')=False
LegacyVersion('0.1.0.windows.2') <  Version('0.2.0')=True
LegacyVersion('0.1.0.windows.2') <= Version('0.2.0')=True
Version('0.1.0') >  LegacyVersion('0.2.0.windows.2')=True
Version('0.1.0') >= LegacyVersion('0.2.0.windows.2')=True
Version('0.1.0') <  LegacyVersion('0.2.0.windows.2')=False
Version('0.1.0') <= LegacyVersion('0.2.0.windows.2')=False
```

@cottsay So to prevent really complex code, I suggest to even require packaging>=22

cottsay: > https://github.com/colcon/colcon-cmake/pull/156#discussion_r2728778851
> But we don't use that feature here.

Sorry, my comment wasn't clear. I'm asking if we can relax it to at least `>=16.8` so that we can continue to support RHEL 8 for Humble without patching. Nothing to do with the feature requiring v17 in colcon-cmake.

> So to prevent really complex code, I suggest to even require packaging>=22

That will make the package unusable on Ubuntu Jammy (21.3), RHEL 8 (16.8) and RHEL 9 (20.9). And that's just REP 2000 platforms - I'm sure there are community platforms out there that would be affected by such a tight constraint as well.

MatthijsBurgh: Ok I will dive into it

MatthijsBurgh:
@cottsay done, please review

MatthijsBurgh:
@cottsay friendly ping

tfoote (approved):
Thanks I think that this is good to go now.

PR #196, opened 2025-09-03
Catch errors thrown by manifest providers

harrysarson:
Distribution objects have a list of manifest providers (both release and source) and will try each in turn until one returns non-null. However, the providers all raise exceptions rather than returning null which means the Distribution object fails to try more than one.

Typically code uses `CachedManifestProvider` (and
`CachedSourceManifestProvider`) which has this try/except logic so the looping over manifest providers works as intended.

This commit copies the try/except logic from these  manifest providers directly into the methods on the Distribution class.

### DIFF
--- a/src/rosdistro/distribution.py
+++ b/src/rosdistro/distribution.py
@@ -37,6 +37,7 @@ from .manifest_provider.github import github_manifest_provider, github_source_ma
 from .manifest_provider.gitlab import gitlab_manifest_provider, gitlab_source_manifest_provider
 from .manifest_provider.tar import tar_manifest_provider, tar_source_manifest_provider
 from .package import Package
+from . import logger
 
 
 class Distribution(object):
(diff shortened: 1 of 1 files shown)


tfoote (commented):
Thanks for digging into this. It looks reasonable. It seems like we might have had a behavior evolution that changed but wasn't resolved for all the use cases. 

If we're considering that this was supposed to degrade gracefully and skip providers that aren't present. Why did you add the exception rethrow versus returning none.  The CachedManifestProvider doesn't use that. 

https://github.com/ros-infrastructure/rosdistro/blob/fad8d9f647631945847cb18bc1d1f43008d7a282/src/rosdistro/manifest_provider/cache.py#L84-L91

harrysarson:
> Why did you add the exception rethrow versus returning none. The CachedManifestProvider doesn't use that.

I was trying to minimise the change in behaviour introduced by my patch. I thought some code might be relying on `get_release_package_xml` throwing an error so wanted to not change that. 

Very happy to switch this to return None though.

---

Looks like the CI failure is a network error: `FAILED test/test_index.py::test_get_index_from_http_with_query_parameters - urllib.error.URLError: <urlopen error <urlopen error timed out> (http://localhost:9876/index_v3.yaml?raw&at=master)>` would it be possible to retry it?

Is the codecov comment something I need to fix? Looking at `Please [upload](https://docs.codecov.com/docs/codecov-uploader) report for BASE (master@fad8d9f). [Learn more](https://docs.codecov.io/docs/error-reference?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ros-infrastructure#section-missing-base-commit) about missing BASE report.` Maybe CI needs to run on the master branch?

src/rosdistro/distribution.py
```diff
@@ -74,10 +75,23 @@ def get_release_package_xml(self, pkg_name):
...
+                except Exception as e:
+                    # pass and try next manifest provider
+                    logger.debug('Skipped "%s()": %s' % (mp.__name__, e))
+                    final_exception = e
+                    continue
+
                 if package_xml is not None:
                     break
+
+            if final_exception is not None:
```
cottsay:
```suggestion
            if final_exception is not None and package_xml is None:
```

cottsay (approved):
Thanks!

PR #198, opened 2025-10-13
Drop debugging output from gitlab manifest provider

cottsay:
This statement was clearly left over from debugging and can be demoted to 'debug' level.

Resolves:
```
src/rosdistro/manifest_provider/gitlab.py:54: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```

Follow-up to #163.


tfoote (approved):
lgtm

PR #200, opened 2025-10-13
Detect http test readiness with retry loop

cottsay:
This replaces the existing static sleep.

### DIFF
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -66,10 +66,26 @@ def test_get_index_v4():
     get_distribution_file(i, 'foo')
 
 
+def _wait_for_server():
+    import socket
+    import time
+    i = 1
+    while True:
+        try:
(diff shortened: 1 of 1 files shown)


cottsay:
I don't really understand what's going on here, but some of the `macos-latest` GitHub Actions runners are taking an extremely long time to run our tests here. Normal runs are on the order of ~10 seconds, but new Python versions on macOS are more like ~90 seconds.

This retry loop should allow other platforms to continue running the tests quickly but will allow for the unpredictable performance we're seeing on macOS.

cottsay:
We're seeing this in two recent PRs:
- https://github.com/ros-infrastructure/rosdistro/pull/198
- https://github.com/ros-infrastructure/rosdistro/pull/171

tfoote (approved):
Thanks this is a better approach. Not sure what's going on with MacOS

